PR #31, opened 2024-12-13
remove undefined coding_sql_where in fhir query

minrk:
This is causing a 500 error getting Observations


s1monj:
Thanks @minrk I missed that one

PR #32, opened 2024-12-14
remove double-quotes from Observation sql

minrk:
quotes [shouldn't be added](https://docs.djangoproject.com/en/5.1/topics/db/sql/#passing-parameters-into-raw) around placeholder fields, causes syntax errors because django already quotes parameters, resulting in:

```
syntax error at or near "core_codeableconcept"
LINE 47:             core_codeableconcept.coding_system LIKE ''*'' AN...
```

follow-up to #31 

I can't test this, but it would be good to verify that the fhir Observation endpoint works when deploying this, in case more patches are needed


s1monj:
Thanks @minrk turns out there were a couple of other sloppy mistakes I had to fix (now that I'm testing against the correct codebase) so I did a fresh commit instead of merging this

Issue #28, opened 2024-12-05
Consider adding script to import seed data


ryanlovett:
I wrote a simple script to import the seed data in a cluster environment, where `psql` was not included the docker container running the web app. It inherits the various `DB_*` environment variables. It might be worth including something similar in `jhe/scripts/`. Since it is part of the app's deployment and not the app itself, I didn't feel confident in creating a PR for it. So just for your consideration...

```python
import os

import psycopg2

# Database connection parameters
db_params = {
    'dbname': os.environ.get('DB_NAME'),
    'user': os.environ.get('DB_USER'),
    'password': os.environ.get('DB_PASSWORD'),
    'host': os.environ.get('DB_HOST'),
    'port': os.environ.get('DB_PORT', 5432)
}

# Connect to the database
conn = psycopg2.connect(**db_params)
conn.autocommit = True
cursor = conn.cursor()

# Read and execute the SQL file
with open('/app/seed.sql', 'r') as sql_file:
    sql_commands = sql_file.read()

    # Execute the SQL commands
    cursor.execute(sql_commands)

# Close the connection
cursor.close()
conn.close()

print("Seed data imported successfully.")
```

s1monj:
good idea and looks great thanks @ryanlovett - will find a spot for it

PR #37, opened 2025-03-21
Add seed.py.

ryanlovett:
This is to address #28. I originally wasn't going to add it to the repo, but I can't think of a better place right now.


s1monj:
Thanks @ryanlovett agreed this is a good place for now

PR #65, opened 2025-04-12
feat/github-issue-8-pagination-studies

travis-sauer-oltech:


jhe/core/models.py
```diff
@@ -499,7 +499,32 @@ class Study(models.Model):
     icon_url = models.TextField(null=True, blank=True)
 
     @staticmethod
-    def for_practitioner_organization(practitioner_user_id, organization_id=None, study_id=None):
+    def for_practitioner_organization(practitioner_user_id, organization_id=None, study_id=None, page=None, pageSize=None):
+
+        # TODO: need to make this as DRY as possible, and re-use in other models.
```
travis-sauer-oltech, line 504: This needs to be DRY. Happens in every model.

jhe/core/models.py
```diff
@@ -520,9 +545,50 @@ def for_practitioner_organization(practitioner_user_id, organization_id=None, st
...
+            """.format(
+                study_sql_where=study_sql_where,
+                organization_sql_where=organization_sql_where,
+                pageSize=pageSize,
+                offset=offset
+            )
         
         return Study.objects.raw(q, {'jhe_user_id': practitioner_user_id})
+    
+    @staticmethod
```
travis-sauer-oltech, line 559: This can be made more DRY, but raw SQL queries will be different regardless, so still a good bulk of code, but things like the temporary class can be transformed into a utility function.

PR #70, opened 2025-04-20
feat/gh-issue-56-set-info-log-level

travis-sauer-oltech:


jhe/core/models.py
```diff
@@ -786,6 +786,8 @@ def for_practitioner_organization_study_patient(practitioner_user_id, organizati
 
       print(f"pageSize: {pageSize}, page: {page}")
 
+      logger.info(f"GH issue #56 sanity check")
```
travis-sauer-oltech, line 789: Verified locally that this is output to the console when navigating to observations tab in UI:
image

PR #243, opened 2026-01-17
fix(jhe-pgd): make patient consents idempotent and improve invitation…

travis-sauer-oltech:
… links

- Make POST/PATCH consents upsert via update_or_create (avoid unique constraint failures)
- Use timezone-aware consented_time
- Build multi-host invitation links including hostname/client_id/code_verifier/code
- Parse CH_INVITATION_LINK_EXCLUDE_HOST as boolean; allow env overrides for PKCE verifier/challenge
- Add requirements.txt and fix runtests.py settings module
- Add regression test ensuring duplicate consent POST is safe


requirements.txt
```diff
@@ -0,0 +1,23 @@
+coverage
```
minrk: As you mentioned, I do like having a plain-pip way to install things here. What I would like to avoid  is a complete _duplicate_ of the same information that's bound to get out of sync. I believe the Pipfile can _consume_ this as the true source, though?

Maybe save this change for another PR, since it's not related to the functionality here?

tests/test_patient_viewset.py
```diff
@@ -62,6 +62,15 @@ def test_patient_practitioner_can_update_own_consents(self):
...
         )
         self.assertEqual(response.status_code, 200)
+
+        # Idempotency: a second POST with the same payload should not create duplicates.
+        response2 = self.client.post(
+            f"/api/v1/patients/{self.patient.id}/consents",
+            data=payload,
+            format="json",
+        )
+        self.assertEqual(response2.status_code, 200)
```
surfdoc: Consider also asserting the count did not increase:
```python
consent_count_before = StudyPatientScopeConsent.objects.count()
response2 = self.client.post(...)
self.assertEqual(StudyPatientScopeConsent.objects.count(), consent_count_before)
```

surfdoc:
@travis-sauer-oltech - is this still relevant or should we close?

travis-sauer-oltech:
@surfdoc Closed out. Aside from being 6-months stale, this was built before the Open Wearable integration into our PGD Web App.

PR #246, opened 2026-01-28
run pre-commit to satisfy linters

minrk:


minrk:
@s1monj @travis-sauer-oltech if you run `pre-commit install` in the top-level of the repo, it'll make sure lint checks always pass. This will also happen if you work via PRs.

PR #276, opened 2026-02-06
Fix pagination for FHIR endpoints

### LINKED ISSUE #204: pagination stops at 1000 entries
When making a paginated request for Observations, the pagination stops at the 1000th row, even when there are many more records. I think this is because a LIMIT is being applied twice, but I'm not sure exactly how.

### LINKED ISSUE #273: fhir list views do not return documented Bundles
### Story Points (or TBE)

TBE

### Effort estimate (hours)

_No response_

### Description

the README suggests that pagination on `/fhir` list endpoints return FHIR Bundle resources:

```python
# GET /fhir/r5/Observation?patient._has:Group:member:_id=30001&patient=40001&code=https://w3id.org/openmhealth|omh:blood-pressure:4.0
{
    "resourceType": "Bundle",
    "type": "searchset",
    "entry": [
        {
            "resource": {
                "resourceType": "Observation",
```

but they actually are:

```python
{'count': 4018,
 'next': 'http://localhost:8000/fhir/r5/Observation?_count=1&_page=2&patient=40006',
 'previous': None,
 'results': [{'resource': {'resourceType': 'Observation', ...
```

### LINKED ISSUE #274: Paginate FHIR endpoints in query
### Story Points (or TBE)

TBE

### Effort estimate (hours)

_No response_

### Description

Pagination of fhir endpoints like Observation, Patient, etc. is current implemented by retrieving all matching entries in `fhir_search`, and then applying pagination to a Python list _after_ retrieving all results.

This is related to #204 where there is an unconditional limit and offset which are never set (limit is always 1000, offset always 0), which are always retrieved, and then the full list is paginated because `get_queryset` returns the complete _list_, not a queryset. So that means an $O(N^2)$ cost to retrieving N results if they don't fit on the page. I believe this led to a number of the errors in our JupyterCon demo.

minrk:
- fixes Bundle schema for fhir pagination endpoints (Patient, Observation) closes #273
- move serialization code from `fhir_search` which is O(total results) to serializers, which is O(page) closes #274
- uses previously unused PaginatedQuerySet class in FHIRPagination to get in-db pagination via django Pagination API
- removes unused AdminModelRawManager
- removes page arguments from `fhir_search`
- adds tests for fhir Observation endpoint

closes #206 
closes #204

### DIFF
--- a/core/serializers.py
+++ b/core/serializers.py
@@ -493,6 +493,14 @@ class FHIRObservationSerializer(serializers.ModelSerializer):
         ]
 
     def to_representation(self, record):
+        # deserialize json fields
+        record.meta = json.loads(record.meta)
+        record.identifier = list(filter(lambda item: item is not None, json.loads(record.identifier)))
+        if len(record.identifier) == 0:
+            del record.identifier
+        record.subject = json.loads(record.subject)
+        record.code = json.loads(record.code)
+        record.value_attachment = json.loads(record.value_attachment)
         as_dict = super().to_representation(record)
         # validate
         try:
@@ -505,20 +513,6 @@ class FHIRObservationSerializer(serializers.ModelSerializer):
 class FHIRBundledObservationSerializer(serializers.Serializer):
     # TBD: full_url = serializers.CharField()
     resource = FHIRObservationSerializer(required=False, read_only=True, source="*")
-
-    def to_representation(self, record):
-        # the raw query in `fhir_search` doesn't deserialize everything
-        # (e.g. jsonb are still strings)
-        # so we need to deserialize some by hand
-        # Extra handling if list can potentially contain nulls
-        record.meta = json.loads(record.meta)
-        record.identifier = list(filter(lambda item: item is not None, json.loads(record.identifier)))
-        if len(record.identifier) == 0:
-            del record.identifier
-        record.subject = json.loads(record.subject)
-        record.code = json.loads(record.code)
-        record.value_attachment = json.loads(record.value_attachment)
-        return super().to_representation(record)
 
 
 class FHIRPatientSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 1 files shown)


minrk:
btw, here's a [quick benchmark](https://gist.github.com/minrk/5cbf106d1cebb7b41ffd41dd614949af) of getting observations, using the fhir and admin APIs, comparing main, this PR, and another branch that I'll open that simplifies admin to follow the pattern here:

Screenshot 2026-02-06 at 17 01 29

The benchmark is run for each of page_size=20, 100, 1000.

You can see the catastrophic performance on main (it also produces incorrect results for N > 1000), which is fixed here.
The admin API is faster than fhir, but not by a ton, and it mostly has to do with how many queries are run (smaller page size), since the count query runs over the full N for each page (no cache).

The simplified admin pagination, which removes the custom count methods, does have a cost, but I think it is one that is small enough that it's worth the simplification.

Noting the scale, this fetches 10k results in ~1 second, where it couldn't do that in minutes before.

I think this also suggests increasing the default page size to at least 100, since the cost of a large number of requests is a good deal higher than the cost of the larger request, which are in the 10s of milliseconds.

core/models.py
```diff
@@ -1339,28 +1292,22 @@ def fhir_search(
             LEFT JOIN core_study ON core_study.id=core_studypatient.study_id
             JOIN core_organization ON core_organization.id=core_study.organization_id
             JOIN core_patientorganization ON core_patientorganization.organization_id=core_organization.id
-            WHERE core_patientorganization.patient_id={patient_user_id}
```
travis-sauer-oltech: @minrk Was this filtration moved to the view layer?

travis-sauer-oltech: I think it would be best to create a test around this to prove that the authorization remains true:

'A user can't see observations from patients they shouldn't have access to.'

Issue #228, opened 2025-12-02
Sample Observation of BP study includes HR instance


scarini:
The example data in the JHE v.0.0.8 includes one study on BP only (ID 30002, scope request: blood pressure) and one study on BP & HR (ID 30001, scope request: blood pressure, heart rate).
However when reviewing Observations from study ID 30002, instances of heart rate are included (see screenshot)
<img width="1358" height="943" alt="Image" src="https://github.com/user-attachments/assets/03a9b008-544c-4600-86e6-b10ab317a976" />

travis-sauer-oltech:
Story Points: 1

Effort estimate (hours): 1

This can be resolved along with: https://github.com/jupyterhealth/jupyterhealth-exchange/issues/193, and https://github.com/jupyterhealth/jupyterhealth-exchange/issues/227

Issue #193, opened 2025-10-11
Don't use real institutions/labs for example data


s1monj:
Currently the test data uses real institutions/labs (eg UCSF). Use fictitious names instead.

s1monj:
@travis-sauer-oltech I think this is just a simple matter of renaming the Organizations in the seed script. Would be good to use some kind of a theme - here are some ideas: https://chatgpt.com/share/695499c5-5ff8-800e-b2f3-462db862792e

Please do a find and replace in the script and the docs when you have a chance

PR #300, opened 2026-02-17
feat(seed): replace real institutions with synthetic names, fix obser…

travis-sauer-oltech:
…vation scope filtering

- renamed all real institution/lab names to fictional Example names (GHI #193/#227)
- fixed seed observation loops to scope consents locally (GHI #228 Bug 1)
- added JOIN core_studyscoperequest to filter observations by study scope (GHI #228 Bug 2)
- removed @pytest.mark.xfail from now-passing test
- added 15 new tests in test_observation_scope_filtering.py
- updated iglu.py and README.md references

### DIFF
--- a/tests/test_observation_scope_filtering.py
+++ b/tests/test_observation_scope_filtering.py
@@ -1,0 +1,435 @@
+"""
+Tests for GHI #228: Observation scope filtering by study
+
+Validates that observations are correctly filtered by study scope codes,
+ensuring that a study requesting only Blood Pressure does NOT surface
+Heart Rate observations (and vice versa), even when the patient is
+enrolled in multiple studies with different scopes.
+
+Also validates the seed command (GHI #193 Bug 1) produces no duplicate
+observations across study groups.
+"""
+
+import pytest
+from django.core.management import call_command
+from django.db import connection
+from django.test.utils import CaptureQueriesContext
+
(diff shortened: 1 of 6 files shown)


tests/test_observation_scope_filtering.py
```diff
@@ -0,0 +1,435 @@
...
+    user.patient.organizations.add(org)
+    return user.patient
+
+
+# ---------------------------------------------------------------------------
+# Unit tests – for_practitioner_organization_study_patient()
+# ---------------------------------------------------------------------------
+
+
+class TestForPractitionerStudyFilters:
```
minrk: is there a reason these are organized as methods on test classes rather than just test functions? The classes don't seem to do anything.

tests/test_observation_scope_filtering.py
```diff
@@ -0,0 +1,435 @@
...
+
+# ---------------------------------------------------------------------------
+# Integration tests – FHIR API endpoint
+# ---------------------------------------------------------------------------
+
+
+class TestFhirApiObservationByStudyScope:
+    """Integration tests hitting the /fhir/r5/Observation endpoint"""
+
+    @pytest.fixture
```
minrk: No reason to reimplement this fixture, I think, since we already have it. Fixtures defined in `conftest` are already available in all tests. You can check the other tests for how these are used.

PR #304, opened 2026-02-23
Feat/smoke tests

travis-sauer-oltech:


minrk (commented):
I think in principle these make sense.

I think most of these tests should be run in the regular test suite, too, is it possible to configure the tests to run against a localhost URL when smoke-url is not specified, and also run against a live instance when specified?

That seems okay as a later improvement, too.

surfdoc:
@travis-sauer-oltech - is this still relevant or should we close?

travis-sauer-oltech:
@surfdoc Closing this as well. 4.5 months worth of staleness. Work has been done in other issues.

Issue #348, opened 2026-04-13
OW POC - syncing Oura heart rate data from S3/MinIO


s1monj:
### Background
- When patient heart rate data arrives in OW it needs to end up in JHE formatted as OMH and packed into an Observation
- All data that arrives in the OW S3 is sequential and ordered by time
- JHE therefore needs to store the OW S3 key of each observation

### TBD
- For now we are using the patient user `identifier` column to save the OW patient ID (we need to improve this later)
- Add a new column to observations called `ow_key` that is a string and is nullable that will look something like this: `oura/api_response/2026-04-10/27b81d5b-c038-4c3a-a395-caf246b7ae79/c1f1a819c47a.json`
  - NB: it's not unique because multiple observations might come from the same file
- Add a new endpoint `GET /api/v1/ow/sync` and when it's called it does the following:
  - Lookup observations.ow_key, sort descending and limit 1
  - Use this ow_key to make a call to MinIO using `StartAfter: {ow_key}` and filtering just for the files used for heart rate
  - Pass each file through the https://github.com/the-commons-project/omh-shim library
  - Write the observation along with the corresponding `ow_key` value

Issue #331, opened 2026-03-30
OW POC - user creation and Oura auth flow


s1monj:
### Story Points (or TBE)

TBE

## Background
- Proof of concept integration with Open Wearables
- No fancy UI for now just log updates to HTML page
- Steps 1-13 in diagram below

### TBD
1. Register a new client for OW in JHE from the UI
2. Add to study and issue link to patient
3. Switch to feat/ow-poc
4. Follow the steps in `core/templates/ow_client/launch.html`
 
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/189192b9-dabb-4604-bd78-559ee0b52001" />

PR #350, opened 2026-04-14
feat: OW PoC user creation and Oura auth flow

travis-sauer-oltech:
Closes out https://github.com/jupyterhealth/jupyterhealth-exchange/issues/331 and https://github.com/jupyterhealth/jupyterhealth-exchange/issues/348.


core/templates/partials/base_ow_client.html
```diff
@@ -0,0 +1,21 @@
...
+    <link
+      rel="shortcut icon"
+      type="image/png"
+      href="{% static 'favicon.ico' %}"
+    />
+    {% block head %} {% endblock %}
+  </head>
+  <body class="h-100">
+    {% block body_content %} {% endblock %}
+    <script src="{% static 'js/common.js' %}"></script>
```
s1monj: @travis-sauer-oltech because these two scripts (common.js and ow-client.js) don't have any top level lines to be called at include-time (ie they're all functions) we can move them inside the html head block - use https://github.com/jupyterhealth/jupyterhealth-exchange/blob/main/core/templates/partials/base.html as a reference

core/static/js/ow-client.js
```diff
@@ -0,0 +1,142 @@
+// ────────────────────────────────────────────────────
+// OW Client - helper functions for the Open Wearables
+// patient integration flow. Uses common.js utilities.
+// ────────────────────────────────────────────────────
+
+var TOKEN_ENDPOINT = window.location.origin + "/o/token/";
+var API_ENDPOINT = window.location.origin + "/api/v1/";
+
```
s1monj, line 8: @travis-sauer-oltech leave these here for now but just as a reminder for later we should move them to core/templates/client/client_settings.js

s1monj (changes requested):
Just one small change to keep src imports consistent while we remember thx

core/static/js/ow-client.js
travis-sauer-oltech: Created a backlog ticket item to track this: https://github.com/jupyterhealth/jupyterhealth-exchange/issues/351

PR #357, opened 2026-04-21
chore: split models into files

s1monj:
- models.py was 14K lines long
- split into individual files


minrk:
@s1monj I would recommend checking the test status before merging, so we don't merge broken things. This caused tests to fail with:

> AttributeError: <module 'core.models' from '/home/runner/work/jupyterhealth-exchange/jupyterhealth-exchange/core/models/__init__.py'> does not have the attribute 'get_setting'

PR #378, opened 2026-05-08
Feat/ow poc

travis-sauer-oltech:


core/services/ow_ingest/raw_payload_reader.py
```diff
@@ -0,0 +1,68 @@
...
+    key: str
+    last_modified: datetime
+
+
+def get_client():
+    """Build a boto3 S3 client from ``ow.s3.*`` JheSettings."""
+    return boto3.client(
+        "s3",
+        endpoint_url=get_setting("ow.s3.endpoint_url", "http://localhost:9000"),
+        aws_access_key_id=get_setting("ow.s3.access_key_id", "minioadmin"),
```
minrk: credentials probably shouldn't have default values

core/services/ow_ingest/raw_payload_reader.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def list_new_objects(ow_user_id: str, since: datetime) -> list[S3ObjectInfo]:
+    """List S3 objects belonging to ``ow_user_id`` newer than ``since``.
+
+    Filters by ``"/{ow_user_id}/"`` substring match on the S3 key (matches
+    Open Wearables' default key layout: ``<prefix>/<provider>/<endpoint>/<user_id>/<...>``).
+    """
+    client = get_client()
+    bucket = get_setting("ow.s3.bucket_name", "raw-payloads")
```
minrk: probably shouldn't have default values for global namespace like like s3 bucket name

doc/open-wearables/README.md
```diff
@@ -0,0 +1,141 @@
...
+| `ow.sync_in_progress` | bool | `false` | Lock auto-managed by `ow_poll`. |
+
+Raw-mode S3 settings (only required when `ow.ingest_mode=raw`):
+
+| Key | Default |
+|-----|---------|
+| `ow.s3.endpoint_url` | `http://localhost:9000` |
+| `ow.s3.access_key_id` | `minioadmin` |
+| `ow.s3.secret_access_key` | `minioadmin` |
+| `ow.s3.bucket_name` | `raw-payloads` |
```
minrk: same here - let's not have any default values for things folks _need_ to set, like key id, key, bucket name, and endpoint_url

Dockerfile
```diff
@@ -20,9 +20,13 @@ RUN --mount=type=cache,target=${XDG_CACHE_DIR} \
     export PIP_CACHE_DIR=$XDG_CACHE_DIR/pip \
  && export PIPENV_CACHE_DIR=$XDG_CACHE_DIR/pipenv \
  && pip install pipenv \
- && pipenv install --system \
+ && pipenv install --deploy --system \
  && pip uninstall -y pipenv
 
+# supercronic for the optional jhe_cron sidecar (runs ow_poll on a schedule).
+ADD https://github.com/aptible/supercronic/releases/download/v0.2.29/supercronic-linux-amd64 /usr/local/bin/supercronic
```
minrk: if we keep this, it must match the image's architecture, not hardcode amd64. But I'm also not quite sure a cron implementation running this from cli is the right solution, but I think it's okay for now.

core/management/commands/ow_poll.py
```diff
@@ -0,0 +1,355 @@
...
+            "--patient-id",
+            type=int,
+            default=None,
+            help="Poll only the specified patient (by Patient.id). For debugging/backfill.",
+        )
+
+    def handle(self, *args, **options):
+        if not bool(get_setting("module.ow", False)):
+            logger.info("ow_poll skipped: module.ow=false")
+            self.stdout.write("ow_poll skipped: module.ow=false")
```
minrk, line 94: why log the same message twice?

core/management/commands/ow_poll.py
```diff
@@ -0,0 +1,355 @@
...
+                self.stdout.write("ow_poll skipped: ow.sync_in_progress=true")
+                return
+            setting.set_value("bool", True)
+            setting.save()
+        cache.delete(f"jhe_setting:{_SYNC_LOCK_KEY}")
+
+        try:
+            self._run_poll(options)
+        finally:
+            _set_sync_lock(False)
```
minrk: If the process crashes during run_poll (i.e. it's killed and this finally never sets the lock to False), what is the recovery process? Is there any indication of when the lock was acquired or what is holding it? That's usually useful information when deciding if a lock is stale and should be forcefully released.

Dockerfile
```diff
@@ -23,6 +23,14 @@ RUN --mount=type=cache,target=${XDG_CACHE_DIR} \
  && pipenv install --system \
  && pip uninstall -y pipenv
 
+# supercronic for the optional jhe_cron sidecar (runs ow_poll on a schedule).
+# TARGETARCH is automatically set by Docker BuildKit / buildx to match the
+# image's target platform (amd64, arm64, arm). Supercronic publishes a
+# matching binary for each.
+ARG TARGETARCH
+ADD https://github.com/aptible/supercronic/releases/download/v0.2.29/supercronic-linux-${TARGETARCH} /usr/local/bin/supercronic
+RUN chmod +x /usr/local/bin/supercronic
```
minrk: Running chmod in a second layer doubles the size of the modified files in the image. You can do this in the ADD command:

```suggestion
ADD --chmod=755 https://github.com/aptible/supercronic/releases/download/v0.2.29/supercronic-linux-${TARGETARCH} /usr/local/bin/supercronic
```

travis-sauer-oltech: @minrk interesting! Thanks for the feedback on this one.

travis-sauer-oltech: It makes sense logically now; such an important little optimization.

core/management/commands/ow_poll.py
travis-sauer-oltech: Not sure why this isn't showing as 'outdated,' but only stdout is present now.

PR #451, opened 2026-05-27
fix lookup of admin ui application

minrk:
first() doesn't raise when none is found, it returns None, leading to no warning in the case when it's not found


travis-sauer-oltech:
@minrk Nice catch, not-found case was silently skipping the warning. the explicit set and check surfaces it and with a more informed message. Merging this.

Issue #452, opened 2026-05-27
new ow tests are not passing


minrk:
### Effort estimate (hours)

2

### Description

The ow tests added in #378 don't pass, so our CI is failing. Let's try to make sure tests are passing _before_ merging any PRs.

We can add merge conditions to PRs if that would help, so it's harder to ignore failing tests.

minrk:
Our CI stops at one failure, but when I run the tests locally, I get 10/16 tests failing:

```
FAILED tests/test_ow_integration.py::TestCreateOwUser::test_creates_user_in_ow - assert 500 == 201
FAILED tests/test_ow_integration.py::TestCreateOwUser::test_returns_existing_ow_id_if_already_linked - assert 500 == 200
FAILED tests/test_ow_integration.py::TestCreateOwUser::test_handles_409_conflict_with_lookup - assert 500 == 200
FAILED tests/test_ow_integration.py::TestCreateOwUser::test_handles_ow_api_error - assert 500 == 502
FAILED tests/test_ow_integration.py::TestCreateOwUser::test_handles_connection_error - assert 500 == 502
FAILED tests/test_ow_integration.py::TestOuraAuthorize::test_no_ow_user_returns_400 - assert 500 == 400
FAILED tests/test_ow_integration.py::TestOuraAuthorize::test_returns_authorization_url - assert 500 == 200
FAILED tests/test_ow_integration.py::TestOuraAuthorize::test_uses_custom_redirect_uri - assert 500 == 200
FAILED tests/test_ow_integration.py::TestOuraAuthorize::test_handles_ow_api_error - assert 500 == 502
FAILED tests/test_ow_integration.py::TestOuraAuthorize::test_handles_connection_error - assert 500 == 502
```

travis-sauer-oltech:
@minrk I like the idea of adding merge conditions to PRs. We're not using the full power of GitHub, and they've made some wonderful additions over the years.

cc: @s1monj @surfdoc

Issue #454, opened 2026-05-28
test_stale_lock_is_force_released is flaky


travis-sauer-oltech:
Passes alone, fails inside the full suite (`pytest tests/test_ow_poll.py::test_stale_lock_is_force_released`). Suggests time-dependent assertion - swap real datetime for an injected clock/freezegun. I suggest using the python test library freezegun because it's less invasive than injecting a clock.

travis-sauer-oltech:
Root cause: test isolation, not timing. `test_unknown_mode_aborts` sets`ow.ingest_mode = "bogus"`, which `get_setting()` caches process-wide; the DB row rolls back between tests but the cache entry doesn't, so this test inherited `"bogus"`, aborted the poll, and saw 0 observations instead of 1 (CI run 26585810323). The freezegun/clock idea doesn't apply. Fix: the test declares its own `ow.ingest_mode = "normalized"` precondition. Closing via the linked PR.

Issue #453, opened 2026-05-28
test_audience_mismatch fails on main


travis-sauer-oltech:
Pre-existing test failure unrelated to #452 . Reproduce: `pytest tests/test_auth.py::test_audience_mismatch`. The mismatched-audience JWT is accepted instead of rejected.

travis-sauer-oltech:
## Root cause:

not a product bug. The RFC 8693 audience check (core/views/common.py) is correct and unchanged since PR #272. The failure came from test fragility: get_setting("site.url", settings.SITE_URL) reads Django's default LocMemCache, which is per-process, not cleared between tests, and not rolled back with the DB transaction. A leaked/ambient site.url equal to the test's "mismatched" audience made the check correctly accept it.

## Chosen path:

fixed the whole class, not just this test.

tests/conftest.py: autouse fixture clears the cache around every test (prevents cross-test leaks for all get_setting users).
tests/test_auth.py: test_audience_mismatch now derives the audience and asserts from settings.SITE_URL, so it's independent of env/.env.

This surfaced a second latent bug: test_observation_pagination's < 8 query-count assertion only passed because a prior test had warmed the cache. Fixed by priming the cache inside that test.
Full suite: 249 passed, no regressions. Closing once merged.

PR #455, opened 2026-05-28
fix(tests): align ow_integration assertions with deployed view contra…

travis-sauer-oltech:
…ct (#452)


travis-sauer-oltech:
@minrk Per #452 , I created #453 and #454 based on issues with tests prior to the original ow-poc PR.

Issue #323, opened 2026-03-20
Icons next to top-level Organization show more actions than allowed


scarini:
### Story Points (or TBE)

TBE

### Description

I logged in as a user with only Viewer privileges for the Organization shown in the screenshot. The icons displayed for the top-level Organization suggest I could delete the Organization. When I tried I got an error message. 
Icons next to the top-level Organization should reflect the actions allowed to the user 
<img width="1427" height="265" alt="Image" src="https://github.com/user-attachments/assets/15268c3d-c9d5-4479-9dcc-428028412385" />

travis-sauer-oltech:
closed per #466

PR #466, opened 2026-05-30
fix: gate top-level organization action icons behind manage permission

travis-sauer-oltech:
Top-level org Update/Create/Delete icons were shown whenever an org was selected, regardless of role, so Viewers saw actions they couldn't perform. Gate them behind canManagePractitionersInOrg, matching the recursive child tree. Read icon stays always visible.

### DIFF
--- a/core/templates/client/client_templates/organizations.html
+++ b/core/templates/client/client_templates/organizations.html
@@ -22,6 +22,7 @@
           read: true
         })"
       }}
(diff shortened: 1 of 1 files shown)


travis-sauer-oltech:
Closes #323 . Tested across:

## Negative results:

viewer_victor and member_megan

## Positive results:

manager_mary, manager_mark, and admin

Issue #235, opened 2025-12-21
Sample CGM data does not comply with IEEE 1752/OMH format standards


surfdoc:
## Background

The sample CGM data in the JupyterHealth Exchange (used in the CGM tutorial) does not comply with IEEE 1752/Open mHealth data format standards. This was reported in jupyterhealth/software-documentation#60.

### 1. Incorrect Unit Format
- **Current**: `blood_glucose_unit: "MGDL"`
- **Should be**: `mg/dL` (lowercase with forward slash per OMH schema)

### 2. Wrong Modality
- **Current**: `modality: "self-reported"`
- **Should be**: `sensed` (CGM devices automatically sense glucose, they are not self-reported)

### 3. Invalid Temporal Relationship to Meal
- **Current**: `temporal_relationship_to_meal: "unknown"`
- **Should be**: One of the valid OMH values: `fasting`, `not fasting`, `before meal`, `after meal`, `2 hours postprandial`, or omitted entirely if truly unknown

### 4. Inadequate External Datasheet Provenance
- **Current**: `external_datasheets_0_datasheet_reference: "Dexcom"`
- **Issue**: Generic manufacturer name without proper device provenance (model number, firmware version)

### 5. Schema Version Inconsistency
- Data shows `schema_id_version: 3.1` but uses code `omh:blood-glucose:4.0`
- Suggests potential schema validation issues

## Impact

1. **Educational**: Users learning from the tutorial will learn incorrect data formats
2. **Validation**: Suggests schema validation may not be properly enforcing IEEE 1752/OMH compliance at ingestion
3. **Data Quality**: Non-compliant data in the database undermines platform credibility

## Source

The sample data appears to be adapted from PLOS Biology research (https://doi.org/10.1371/journal.pbio.2005143.s010) which was not originally in OMH format.

## Requested Actions

1. **Investigate** why OMH-non-compliant data exists in the Exchange
   - Review data ingestion process
   - Check where validation is (or isn't) happening

2. **Strengthen schema validation** to reject improperly formatted data
   - Enforce proper unit formats (`mg/dL` not `MGDL`)
   - Validate modality values against OMH spec
   - Validate temporal_relationship_to_meal against allowed values
   - Require proper device provenance in external_datasheets

3. **Generate properly formatted sample data**
   - Create new sample CGM data that fully conforms to IEEE 1752 v4.0
   - Or fix the existing sample data in the database to be compliant

[...]

PR #494, opened 2026-06-04
fix(omh): vendor missing blood-glucose $ref schemas and enforce CGM v…

travis-sauer-oltech:
…alidation (#235)


surfdoc (approved):
Looks Good! Thank You.

Issue #192, opened 2025-10-11
Infinite redirect to sign-in with OAuth error


s1monj:
If an OAuth error is returned form the server (eg invalid private key) the front end retries sign-in indefinitely. Instead the error should be relayed to the front end.

To recreate, set OIDC_RSA_PRIVATE_KEY to empty key in `.env`

PR #496, opened 2026-06-05
fix: relay OAuth errors to front end instead of infinite sign-in retr…

travis-sauer-oltech:
…y (#192)


travis-sauer-oltech:
closes #192 .

Issue #498, opened 2026-06-05
Investigate jhe.fly.dev 500 server error


travis-sauer-oltech:
Per @minrk [fly.dev](http://fly.dev/) deployment is failing during db migration:

2026-06-04T17:34:49Z Running migrations:
2026-06-04T17:34:49Z   Applying core.0030_remove_jhesetting_setting_id...ALTER TABLE "core_jhesetting" DROP CONSTRAINT "core_jhesetting_unique_key_setting_id"; (params None)
2026-06-04T17:34:49Z ALTER TABLE "core_jhesetting" DROP COLUMN "setting_id" CASCADE; (params None)
2026-06-04T17:34:49Z ALTER TABLE "core_jhesetting" ADD CONSTRAINT "core_jhesetting_key_e08298e2_uniq" UNIQUE ("key"); (params None)
2026-06-04T17:34:49Z Traceback (most recent call last):
2026-06-04T17:34:49Z   File "/usr/local/lib/python3.11/site-packages/django/db/backends/utils.py", line 103, in _execute
2026-06-04T17:34:49Z     return self.cursor.execute(sql)
2026-06-04T17:34:49Z            ^^^^^^^^^^^^^^^^^^^^^^^^
2026-06-04T17:34:49Z psycopg2.errors.UniqueViolation: could not create unique index "core_jhesetting_key_e08298e2_uniq"

travis-sauer-oltech:
## Resolved

<b>Root cause</b>: migration `0030_remove_jhesetting_setting_id` made `JheSetting.key` globally unique but never removed the per-client `client.invitation_url` rows that `0027` had already backfilled into `JheClient`. Fresh seed DBs have no such rows, so it passed locally and failed only on prod's real data with `UniqueViolation: Key (key)=(client.invitation_url) is duplicated`.

<b>Fix</b>: added a `RunPython` step to `0030` that deletes the superseded `client.invitation_url` settings before the unique constraint is applied. PR: `fix/ghi-498-jhesetting-0030-duplicate-key`.

PR #500, opened 2026-06-05
fix: drop superseded client.invitation_url settings in migration 0030

travis-sauer-oltech:
Migration 0027 backfilled per-client invitation URLs into JheClient. 0030 makes JheSetting.key globally unique but left the old per-client rows in place, breaking the unique index on databases with real client data (prod). Delete them before applying the constraint.


travis-sauer-oltech:
closes #498

Issue #508, opened 2026-06-05
generate codeable concepts from schemas


minrk:
### Description

Is there a reason codeable concepts must be defined by hand and separately for each deployment, rather than loaded automatically from schemas?

It seems like we should be able to populate these automatically from the list of available schemas.

minrk:
Related: #509 loads all ieee, ieee metabolic draft, and omh schemas automatically, rather than a hand-selected, sometimes edited subset. I think it would be easy to add _all_ schemas, but not all schemas are really codeable concepts, but _most_ are. I'm not sure if there's a clear, general, automatic way to distinguish e.g. `blood-pressure` from `part-of-day`

minrk:
Another reason to possibly not do this is #215 - the codeable concepts should __not__ map 1:1 with schemas, but rather be 1:many for recognized version history of schemas. For example, CodeableConcept (insofar as it represents patient permissions) should be 'heart rate' which should map to a _list_ of schemas to accept, e.g. `omh:heart-rate:1.1,omh:heart-rate:2.0`, which should be modifiable on the deployment side without updating the patient permissions. The _matching_ schema should be in the database, but so should the codeable concept.

s1monj:
@minrk agreed we should remove version from the Code

So we could have codeable_concepts:
```
  id   |        coding_system         |        coding_code        |            text            
-------+------------------------------+---------------------------+----------------------------
 50004 | https://w3id.org/openmhealth | omh:heart-rate            | Heart Rate
 50009 | http://hl7.org/fhir/         | QuestionnaireResponse     | FHIR QuestionnaireResponse
```

Then we can have an OMH schemas table:
```
id | codeable_concept_id | version | schema (json blob)
```

s1monj:
@minrk @surfdoc also not sure how far/abstract we want to take this?

Eg: if there's a future use case of "I want to share all my Heart Rate data" which includes both OMH data and clinical data from the EHR (eg MyChart) then we need to separate the grant/permission/consent codes from the Observation codes.

We would use `http://loinc.org|8867-4` as the universal code for Heart Rate and then associate the OMH code for Heart Rate Observations with the grant code.

```
"code": {
  "coding": [
    {
      "system": "http://loinc.org",
      "code": "8867-4",
      "display": "Heart rate"
    },
    {
      "system": "https://w3id.org/openmhealth",
      "code": "omh:heart-rate",
      "display": "Open mHealth heart rate"
    }
  ]
}
[...]

PR #509, opened 2026-06-05
vendor schemas directly from omh, ieee

### LINKED ISSUE #216: include all of openmhealth
We should probably include the whole openmhealth schema repo in `data/omh/json-schemas` rather than selecting only some files.

We should also perhaps automatically load codeable concepts from available schemas, rather than needing to do that separately.

The current draft version we've been working with is currently at https://opensource.ieee.org/omh/1752-2/-/tree/1.1-metabolic I believe (cc @scarini)

Related to #210

### LINKED ISSUE #443: Vendored schemas drift from canonical OMH + IEEE 1752.1 sources
While integrating [omh-shim](https://github.com/jupyterhealth/omh-shim) with JHE end-to-end, I found that schemas vendored under `data/omh/json-schemas/` differ from the canonical sources at [openmhealth/schemas](https://github.com/openmhealth/schemas) and [opensource.ieee.org/omh/1752](https://opensource.ieee.org/omh/1752). The differences aren't cosmetic — for at least one schema (`omh:sleep-episode:1.1`), the divergence causes OMH-conformant data to be rejected by JHE.

## Evidence: `omh:sleep-episode:1.1`

Three distinct schemas exist under what should be the same id:

**OMH canonical `sleep-episode-1.1.json`** (`openmhealth/schemas@36078a89`):
```json
"sleep_maintenance_efficiency_percentage": {
  "allOf": [{ "$ref": "unit-value-1.x.json" }, ...]
}
```
Expects `{value: 92.5, unit: "%"}`.

**IEEE 1752.1 canonical `sleep-episode-1.0.json`** (`opensource.ieee.org/omh/1752@1.0.2`):
```json
"sleep_maintenance_efficiency_percentage": {
  "$ref": "#/definitions/percent_unit_value"
}
```
Still a structured value (`percent_unit_value`).

**JHE's `data/omh/json-schemas/data/schema-omh_sleep-episode_1-1.json`** (current):
```json
"sleep_maintenance_efficiency_percentage": {
  "type": "number", "minimum": 0, "maximum": 100
}
```
Bare number — matches **neither** canonical source. Only 3 properties total; the rest of the OMH/IEEE schema is missing.

## Practical effect

omh-shim's `convert(source='oura_raw', data_type='sleep_episode', ...)` produces output that validates against canonical OMH (the shim's vendoring target) but FAILS JHE's validation:

```
ValidationError: {'value': 92.5, 'unit': '%'} is not of type 'number'
```

Observable today on omh-shim v1.0.2 against a fresh JHE clone.

## Broader pattern

`diff` shows all 6 vendored OMH body schemas under `data/omh/json-schemas/data/schema-omh_*.json` differ from their canonical counterparts at `openmhealth/schemas`. Most differences happen not to affect validation outcomes for the shapes the shim currently emits — coincidence, not design. Only `sleep_episode` produces an observable rejection with current shim outputs, but the drift is systemic.

## Suggested fix

Adopt the pattern of schema vendoring from canonical sources at a pinned ref, rather than carrying hand-modified copies.

- Schemas with an IEEE 1752.1 version → vendor from `opensource.ieee.org/omh/1752` (e.g., `sleep-episode-1.0.json`)
- Schemas only in OMH → vendor from `openmhealth/schemas` (e.g., `heart-rate-2.0.json`)
- Record the pinned ref in a README so refreshes are reproducible

omh-shim is implementing this pattern in [jupyterhealth/omh-shim#8](https://github.com/jupyterhealth/omh-shim/issues/8) (pin refresh script) and [jupyterhealth/omh-shim#9](https://github.com/jupyterhealth/omh-shim/issues/9) (two-source vendoring). The same pattern applied here would close the drift and make both repos validate against the same schemas.

## Minimum action

Even before scripting, replacing `schema-omh_sleep-episode_1-1.json` with either the canonical OMH or canonical IEEE 1.0 version would unblock sleep_episode uploads from omh-shim.

@travis-sauer-oltech assigning to you for triage.

minrk:
avoid all local modifications, these files should never need to be touched by hand again. All updates should go through `scripts/pull_schemas.py`, and any changes should be done via contributions to upstream repos

removes all path-based lookups, all lookups go through the registry once it is constructed

closes #443
closes #216


core/models/observation.py
```diff
@@ -284,8 +283,8 @@ def _omh_payload(fhir_observation, user_patient):
 
     @staticmethod
     def validate_outer_schema(instance_data):
-        for name in ("data-point-1.0.json", "data-series-1.0.json"):
-            schema = json.loads((settings.DATA_DIR_PATH.schemas_metadata / name).read_text())
+        for code in ("ieee:data-point:1.0", "ieee:data-series:1.0"):
+            schema = code_to_schema("ieee:{name}")
```
surfdoc: This looks like a typo. `"ieee:{name}"` is a plain string (not an f-string), and the loop variable `code` is never used — so both iterations call `code_to_schema("ieee:{name}")`, which hits `ns, name, version = code.split(":")` with only two parts and raises `ValueError` (not caught by the `except ValidationError`). I think it's just meant to be `code_to_schema(code)`.

No live impact today since `validate_outer_schema` has no callers, but it's broken the moment it's wired up.

core/utils.py
```diff
@@ -26,26 +27,45 @@ def __call__(self, uri: str):
...
-    reg = _preload_dir(reg, "", settings.DATA_DIR_PATH.schemas_utility)
-    reg = _preload_dir(reg, ieee_base, settings.DATA_DIR_PATH.schemas_metadata)
-    reg = _preload_dir(reg, ieee_base, settings.DATA_DIR_PATH.schemas_utility)
-    reg = _preload_dir(reg, omh_base, settings.DATA_DIR_PATH.schemas_utility)
-    reg = _preload_dir(reg, omh_utility, settings.DATA_DIR_PATH.schemas_utility)
-
+    for name in ("omh", "ieee"):
+        schema_base = schema_base_uris[name]
+        reg = _preload_dir(reg, schema_base, settings.DATA_DIR_PATH.schemas_dir / name)
+    reg.crawl()
```
surfdoc: `referencing.Registry` is immutable — `crawl()` returns a *new* registry (`return evolve(self, ...)`) rather than mutating in place, so this line is a no-op and the crawled registry is discarded. Did you mean `reg = reg.crawl()`?

Top-level lookups and the blood-glucose test still pass without it (lazy resolution plus the manual self-`$id` registration just above), so the happy path isn't broken — but any eager subresource/anchor discovery intended here isn't actually happening.

core/utils.py
```diff
@@ -26,26 +27,45 @@ def __call__(self, uri: str):
...
 
 
+schema_base_uris = {
+    "ieee": "https://w3id.org/ieee/ieee-1752-schema",
+    "omh": "https://w3id.org/openmhealth/schemas/omh",
+}
+
+
+@lru_cache
+def code_to_schema(code: str) -> str:
```
surfdoc: Nit: return type is annotated `-> str` but the function returns a `dict` (`{"$ref": ...}`).

surfdoc (commented):
Thanks for vendoring these — the single-registry approach is a nice cleanup, and routing every lookup through `code_to_schema`/the registry reads much better than the old path-mangling. A couple of real issues plus one nit inline. None of these block the happy path (CI is green and the live validation path resolves fine), but the first two look like genuine bugs worth fixing before merge.

core/models/observation.py
minrk: I checked, and the last call to this method appears to have been removed in 5980a42dfe378155f6b5e4933623a9dd88afac43

so I'll just drop the method since it clearly has no effect if it doesn't work.

minrk (commented):
Thanks for the review! I removed the unused method and fixed the other issues in utils.

scarini:
Agree with avoiding local modifications of schemas. Overall, the schemas in IEEE 1752 are pretty stable. The schemas in P1752-2 will go through the IEEE ballot process and they may change. Also, there are additional schemas to be added to the 1752-2 repository, so there will be changes in that respect (alert, one of those is blood-glucose). The OMH repository needs some cleaning up, a process I can resume now that the validator and validation process is working again (thank you @minrk !). The end result of the process is to ensure all schemas 1) have $id specified using a w3id reference, 2) they reference IEEE schemas where appropriate and 3) are deprecated when appropriate. 
Deprecated schemas should be excluded from the process described here (see this as an example https://www.openmhealth.org/documentation/#/schema-docs/schema-library/schemas/omh_physical-activity ) 
Changes to apply 2) require either a minor version or a major one

scarini:
A question to me is whether there are schemas with the same name in two repositories: I don't think so but will need to confirm. Another question to me is the way schemas reference other schemas: in updating the OMH schemas I propose to use the same format I started using (see this as an example https://github.com/openmhealth/schemas/blob/main/schema/omh/diastolic-blood-pressure-1.1.json ) A question from me is how best I can go through the review process so I don't break anything downstream.

minrk:
Schemas changing is fine - it's just that we should match whatever is current in the upstream repos. Whenever there's a change, we should update here, which is now scripted (though not _triggered_ automatically, which we can do later, but I don't think that's pressing). That's already happened twice since I opened this PR - OMH fixed the body-weight version, and the IEEE metabolic draft fixed food-entry, both of which are reflected in this repo now.

When there's two schemas with the same name, that's also fine because they exist in different namespaces. When a subschema is referenced by name alone, it will resolve locally (i.e. a reference to `value.json` in an omh schema will resolve to `value.json` in omh, and not any ieee schemas of the same name _unless_ they are referenced by absolute id url, in which case they will also work correctly, as intended.

Examples:

- `physical-activity.json` is not a valid schema to reference—you _must_ specify if it's `omh:physical-activity:1.0` or `ieee:physical-activity:1.0`, so it is never ambiguous which schema is used at the top level.
- _within_  `omh:physical-activity:1.0`, a relative reference to `"$ref": "time-frame-1.x.json"` unambiguously resolves within the same namespace to `omh:time-frame:1.x`, and cannot be resolved to something in ieee unless the ref is absolute, like `"$ref": "https://w3id.org/ieee/ieee-1752-schema/time-frame.json"`. This is a feature of URLs in general and how they are defined in the JSON Schema `$id` field, not anything specific to our implementations.


> A question to me is whether there are schemas with the same name in two repositories: I don't think so but will need to confirm.

There are, but they are almost exclusively metadata field sub-schemas and not an issue. That should be fine, because they will be referenced differently. There's no issue with `omh/physical-activity-1.0` and `ieee/physical-activity-1.0` both existing because they are referenced absolutely, not just by name. There is something of an issue for which schema (_if_ they differ) a given _deployment_ uses to validate, but because they are always referenced _explicitly_ in JHE, there is no actual ambiguity or problem created by loading both under different ids. Which one is chosen happens at a higher level than the code here - where a deployment _chooses_ to define Codeable Concepts, and which schemas are used (#508 is related, but there aren't plans to address that, yet).

> Another question to me is the way schemas reference other schemas

As long as schemas outside the repo you're editing (e.g. omh referencing ieee) are referenced by their absolute w3id URL, everything should work perfectly, and that's precisely what you've linked to, so I think you're on track. That should work great.

> A question from me is how best I can go through the review process so I don't break anything downstream.

The way the omh validator already handles references to ieee should work just fine and match what happens in JHE or ~any other validator. So if the OMH validator continues to pass and resolve references to IEEE, that should be an adequate signal that it should resolve just the same in JHE.

I think pretty much the only way you could break something in JHE without also breaking OMH validation itself is to remove a schema that's actively in use, or _change_ an existing schema that's already in use in an incompatible way without changing its version number. Changing how sub-schemas resolve (e.g. referencing  ieee/unit-value instead of omh/unit-value like in your example) won't break anything.

scarini:
Thank you @minrk for  your answers. The overall approach I am using is: if I am making editorial changes to a schema (e.g., add the $id) no change to version #; if I am changing a reference from a local schema to an IEEE one that is basically the same (e.g., unit-value schema) then I increase the minor version; if I am changing a reference from a local schema to an IEEE one that is not backward compatible, then I increase the major version (this will address the possible cause of a break); if I am deprecating a schema, then I only add the blurb at the top (e.g., sleep-episode) and no change to version #. My most recent PRs all refer mostly to editorial changes and deprecations. I'll let you know when I make changes affecting version numbers

minrk:
Should we merge this, or is there a reason not to do this yet?

PR #618, opened 2026-06-18
chore: make Observation.data_source optional in the Django admin (#603)

### LINKED ISSUE #603: Make Observation.data_source optional in the Django admin (null in DB but form-required)
## Background

Found while testing #527 in the Django admin Add Observation form.

## Context

`core/models/observation.py` defines:

```python
data_source = models.ForeignKey("DataSource", on_delete=models.SET_NULL, null=True)
```

The column is nullable, and the API and normal ingestion often leave it null. But the field does not set `blank=True`, so `blank` defaults to `False` and the Django admin Add/Change form marks `Data source` as required. You cannot save an Observation through the admin without picking a data source, even though the model and DB allow none.

## Proposed

Add `blank=True` to the `data_source` field (validation only; the generated migration is a no-op at the DB level), or mark the field not-required in `ObservationAdmin`. Pick whichever matches the repo pattern.

## Acceptance Criteria

- Saving an Observation in the Django admin with `Data source` left blank succeeds and stores null.
- No DB schema change beyond a no-op migration if `blank=True` is used.

## Out of Scope

Any change to how the API or ingestion sets `data_source`.

travis-sauer-oltech:
data_source is nullable and usually null, but lacked blank=True so the admin form required it. Adds blank=True (no-op AlterField migration) + a regression test. Closes #603.


travis-sauer-oltech:
closed as a duplicate

PR #620, opened 2026-06-18
feat: surface the dashboard access token on the Profile page (#245)

### LINKED ISSUE #245: Improve retrieval of token from dashboard
### Story Points (or TBE)

1

### Effort estimate (hours)

3

### Description

JupyterHealth Client (and other API usage) use a JHE Access token to talk to the API. When not launched via JupyterHub and/or other OAuth launch, users need to pass a token manually. This access is available in the dashboard on the debug page if you click 'Get user', but it would be nice if it were in a more official/discoverable place. Perhaps a User and/or Account page with a 'get my token' button.

Questions:

- are tokens singletons per user, or can a user have multiple access tokens to manage?

travis-sauer-oltech:
Adds a 'Dashboard Access Token' section to the Profile page so any user can copy their current session token for manual API use (previously only on the superuser Debug page). Closes #245.

### DIFF
--- a/core/static/clients/jhe-admin/js/client-jhe-admin.js
+++ b/core/static/clients/jhe-admin/js/client-jhe-admin.js
@@ -2333,6 +2333,22 @@ async function deletePractitionerClient(id) {
   if (await apiRequest("DELETE", `practitioner_clients/${id}`)) await navReturnFromCrud();
 }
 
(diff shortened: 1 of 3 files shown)


travis-sauer-oltech:
minor convenience. since we're aiming for simplicity, closing this as a duplicate to #596.

PR #646, opened 2026-06-28
EHR launch → JHE SSO: id_token-based token exchange (cross-vendor SMART on FHIR)

surfdoc:
# EHR launch → JHE SSO: id_token-based token exchange

## Summary

Lets a provider who launched a SMART on FHIR app from their EHR obtain a JupyterHealth
Exchange access token **without a second login**. The EHR-issued OIDC `id_token` is
exchanged (RFC 8693) for a JHE token: the id_token is verified offline against the EHR's
published JWKS, the `fhirUser` claim is mapped to a JHE Practitioner, and a standard JHE
token is issued.

This replaces the previous token-exchange path, which resolved the subject by calling the
IdP's `userinfo`/`introspection` endpoints at runtime. The new path is offline,
cross-vendor (relies only on ONC g(10)–mandated capabilities), and works across EHR
vendors (e.g. MedPlum, Epic) without per-vendor round-trips.

## We did NOT write our own OAuth / crypto

This is intentionally a thin glue layer over mature, standard libraries:

- **OAuth token issuance** — `django-oauth-toolkit` (`OAuth2Validator.save_bearer_token`,
  `AccessToken`, `TokenView`). Unchanged from before.
- **JWT / JWKS signature verification** — **PyJWT** (`jwt.PyJWKClient`, `jwt.decode`). No
  hand-rolled crypto.

## What we wrote

- **`core/oidc_verify.py`** (new): SMART/OIDC discovery (`.well-known/smart-configuration`,
  falling back to `.well-known/openid-configuration`), JWKS-based `id_token` verification,
  and `fhirUser` → (resource_type, id) parsing.
- **`core/views/common.py`** — `token_exchange` rewritten to the id_token flow:
  trusted-issuer allow-list check, `verify_id_token`, require `fhirUser` is a
  `Practitioner`, map to `JheUser`, issue JHE token.

## Security hardening

- `ALLOWED_ALGS` pinned to asymmetric only (`RS256`, `RS384`, `ES384`) — explicitly never
  `none`, never HS\* (blocks algorithm-confusion attacks).
- Required claims enforced: `exp`, `iat`, `iss`, `aud`.
- Clock-skew `leeway` (30s) on `exp`/`iat`/`nbf` so brief EHR↔JHE drift doesn't cause
  intermittent 401s.
- `jwks_uri` must be `https://`.
- Issuer derived from the (signature-unverified) token's own `iss` so the value passed to
  verification matches the token exactly (handles trailing-slash issuers like MedPlum's
  `https://api.medplum.com/`); allow-list membership compared slash-insensitively.
- Error messages don't leak `kid`.

## Error handling & observability

- Status mapping: untrusted issuer → **403**; bad signature / expired / disallowed alg →
  **401**; non-Practitioner `fhirUser` → **403**; unknown or duplicate practitioner →
  **404**; unconfigured server → **500**; JWKS discovery failure → **502**.
- A trusted issuer returning a non-JSON discovery document yields a clean **502** rather
  than an unhandled 500.
- A duplicate `JheUser.identifier` row (the field has no unique constraint) returns **404**
  and logs, instead of surfacing `MultipleObjectsReturned` as a 500.
- Audit trail: `info` log on successful token issuance (Practitioner + issuer); `warning`
  on untrusted-issuer rejection.

## Design decisions (reviewer notes)

- **Identity maps on the bare `fhirUser` reference, not an issuer-scoped composite.** Each
  health system runs its own JHE instance trusting exactly one EHR, so there is no second
  issuer that could assert another's `Practitioner/<id>`. A composite `(iss, id)` key would
  be gold-plating under this deployment model.
- **NPI was considered as the mapping key and dropped.** JHE has no automatic source for a
  provider's NPI (it isn't in MedPlum/Epic id_tokens by default), so it would require manual
  entry per provider (doesn't scale) or a live FHIR dereference (cross-repo, network-bound).
  Revisit only if a single instance is ever pointed at multiple EHRs.

## New dependency

- `pyjwt` → `pyjwt[crypto]` (adds the `cryptography` backend so PyJWT can verify
  RS\*/ES\* signatures). No other libraries added.

## Configuration

New settings (see `.env.example` / `README.md`), read via `get_setting()` so they can be
overridden at runtime, defaulting to env:

- `TRUSTED_TOKEN_ISSUERS` — allow-list of trusted EHR issuer URLs (the id_token `iss`).
- `TRUSTED_TOKEN_AUDIENCE` — expected `aud` for the id_token (the SMART app's `client_id`).

(Replaces the old `TRUSTED_TOKEN_IDP` single-issuer setting, which is still honored and
folded into the list.)

## Backward compatibility / breaking changes

No existing authentication path is affected. Changes are confined to the `token_exchange`
endpoint; the standard OAuth2 flow (`/o/token/`, `JheTokenView`), session/login auth, the
`OAUTH2_PROVIDER` config, and DRF auth classes are all untouched.

**This affects only the EHR-launch SSO path.** The third-party data API (FHIR/REST push and
pull) authenticates via the standard OAuth2 flow (`/o/token/`, registered clients) and is
unaffected — data vendors integrating against it need no changes.

**One contract change to the token-exchange endpoint itself:** callers must now submit an
EHR-issued **`id_token`**, i.e. `subject_token_type` =
`urn:ietf:params:oauth:token-type:id_token` (previously `...:access_token`). The old
`userinfo`/`introspection`-based resolution has been removed. Any client relying on the
previous access-token exchange behavior must switch to passing an id_token. The `iss` form
field is no longer required — the issuer is taken from the id_token itself.

- Settings are additive and backward-compatible: the legacy `TRUSTED_TOKEN_IDP` is still
  honored (folded into `TRUSTED_TOKEN_ISSUERS`); no setting was removed or repurposed.
- The old `test_auth.py` introspection-flow tests no longer apply; they're replaced by 24
  focused tests across three files — a net coverage gain.

## Diff stats

9 files changed, +520 / −329, 13 commits. The −329 is inflated by reorganizing
`tests/backend/test_auth.py` (−230) into two new focused test files.

| File | Change |
|------|--------|
| `core/oidc_verify.py` | +88 (new) |
| `core/views/common.py` | rewritten token_exchange |
| `tests/backend/test_token_exchange.py` | +162 (new) |
| `tests/test_oidc_verify.py` | +154 (new) |
| `tests/backend/test_auth.py` | −230 (reorganized) |
| `README.md` | +20 |
| `jhe/settings.py` | +14 |
| `.env.example` | +9 |
| `Pipfile` | pyjwt → pyjwt[crypto] |

## Testing (24 tests)

- `tests/test_oidc_verify.py` (13) — discovery + `https`-only and non-JSON rejection; JWKS
  verification; `alg:none`, `HS256`, and wrong-key rejection; audience / issuer / expiry
  checks including within-leeway (accepted) vs beyond-leeway (401); malformed `fhirUser`.
- `tests/backend/test_token_exchange.py` (9) — end-to-end flow, trailing-slash issuer,
  untrusted issuer, expired token, non-Practitioner, unknown and duplicate practitioner,
  audience mismatch.
- `tests/backend/test_auth.py` (2) — missing-argument and wrong-subject-token-type guards.


surfdoc:
@s1monj — could you review this when you have a chance?

It swaps the token-exchange endpoint's identity check from runtime `userinfo`/`introspection` (not guaranteed across EHR vendors) to **offline `id_token` verification against the EHR's JWKS** (PyJWT's `PyJWKClient` — no hand-rolled crypto), mapping the `fhirUser` claim to a JHE Practitioner. Full rationale and the breaking-change scoping (EHR-launch SSO path only — the third-party data API via standard OAuth2 `/o/token/` is untouched) are in the PR description.

Best places to focus: `core/oidc_verify.py` (signature / alg allow-list / iss / aud / exp) and the rewritten `token_exchange` in `core/views/common.py`. 24 tests included. Still pending live MedPlum validation before merge, but ready for eyes. Thanks!

surfdoc:
Branch updated: this PR now contains the full **confidential-client rework** discussed in Slack, so it supersedes the original open-endpoint implementation it previously held:

- `0d4c3bf` — @s1monj's client-auth rework (client authenticates as the seeded "SoF EHR Launch" confidential Application; issued tokens bound to user **and** application; config moved to `auth.sof.*` JheSettings) — cherry-picked from `ehr-launch-jhe-sso-with-client`
- `159b5f9` — fix: HTTP Basic client auth used the wrong header key and could never succeed (regression tests for both credential placements)
- `352124f` — docs/seed: the trusted issuer is the id_token's literal `iss` (Epic: `…/interconnect-fhir-oauth/oauth2`, **not** the FHIR base; MedPlum is the outlier)

**Validation:** end-to-end EHR launches against **MedPlum (hosted)** and the **Epic sandbox (hosted Hyperspace simulator)** both pass with this exact code — same branch, only settings changed between vendors. Epic audit line:

```
Token exchange: issued JHE token for Practitioner 'evNp-KhYwOOqAZn1pZ2enuA3'
from issuer https://fhir.epic.com/interconnect-fhir-oauth/oauth2 to client 'sof-ehr-launch'
```

Full test suite: 471 passed / 28 skipped. Companion PRs: jupyterhealth/jupyterhealth-sof-provider-template#16 (reference app) and jupyterhealth/software-documentation#107 (docs — gives `TOKEN_EXCHANGE_TMP_README.md` its permanent home, after which that file can be deleted here or in a follow-up).

PR #671, opened 2026-07-23
feat(patient-access): SMART 2.2 branding + R4 ingestion (EHR-agnostic)

travis-sauer-oltech:
## What
EHR-agnostic **Patient Access** client (SMART App Launch standalone patient launch): the
patient searches and picks their hospital, is sent to that hospital's SMART login, and
their records are converted R4->R5 and stored in JHE.

## Branding (SMART App Launch 2.2)
- Vendor-neutral `EhrBrand` / `EhrBrandLocation` models + migration `0042`.
- `import_ehr_brands` parses a SMART user-access Brands Bundle; a curated sample is seeded.
- `GET /api/v1/patient-access/brands` search + hospital picker UI; selection authorizes
  against the brand's `fhir_base_url` (iss); SMART `.well-known` discovery resolves endpoints.
- Extensible: no vendor-specific columns; add an EHR by importing its bundle (one future
  change, per-vendor `client_id`, for simultaneous multi-vendor auth).

## Ingestion
- Writes route through `/fhir-import/R4/<type>` (R4->R5 convert, then create), not
  `/FHIR/R5/` directly. Demographics single read fixed (plain instance read).

## Engine (cross_version)
- Primitive choice/extension keys use the map's declared FHIR type (`deceasedBoolean`, not
  `deceasedBool`). Reviewed and approved by Simon.

## Naming
- MyChart -> Patient Access across module, URLs, UI, and tests (EHR-agnostic, per team feedback).

## Testing
- Backend + frontend suites green (models, importer, brands API, picker + pull JS,
  boolean-choice regression, plus the cross-version conformance suite). Verified live against
  the Epic sandbox: Demographics, Medications, Allergies, Labs and 2/7 Conditions save; the
  other 5 Conditions lack `clinicalStatus` in Epic's sandbox data (correctly rejected).

## Note on commits
Includes the FHIR R4->R5 import engine commits (Simon's `feat/fhir-r4-import-patches`) as a
dependency, rebased cleanly onto `main`. @Simon for the engine + the choice-key fix he approved.

## Docs
In `software-documentation`: `jhe/patient-access-integration.md`.


core/static/clients/patient-access/js/client-patient-access.js
```diff
@@ -0,0 +1,333 @@
...
+    return false;
+  }
+  paStoreToken(tokens.access_token);
+  out.textContent += "\nJHE access token received";
+  return true;
+}
+
+// Attach the Epic patient id to the JHE patient (additive). Returns true on success.
+async function paSavePatientIdentifier(jheToken, system, value) {
+  var response = await fetch(API_ENDPOINT + "patient_access/identifier", {
```
surfdoc: **Blocker — URL won't resolve.** This posts to `/api/v1/patient_access/identifier` (underscore), but the route is registered with a hyphen: `api/v1/patient-access/identifier` (core/urls.py:91). Django path matching is exact, so this 404s, `response.ok` is false, and the callback aborts with "failed to store patient id."

Same issue on the brands call at line 179. Suggest switching both to hyphens to match `urls.py`.

core/static/clients/patient-access/js/client-patient-access.js
```diff
@@ -0,0 +1,333 @@
...
+    var ok = await paWriteResource(jheToken, sourceId, pull.type, resource);
+    if (ok) written++;
+    else failed++;
+  }
+  return { written: written, failed: failed, error: null };
+}
+
+// Search hospital brands for the picker. Returns an array of facility rows (or []).
+async function paSearchBrands(jheToken, query) {
+  var url = API_ENDPOINT + "patient_access/brands?q=" + encodeURIComponent(query || "");
```
surfdoc: Same hyphen/underscore mismatch as line 76: this hits `/api/v1/patient_access/brands` but the route is `api/v1/patient-access/brands` (core/urls.py:90). Result is a 404 that `paSearchBrands` swallows into `[]`, so the picker always shows "No hospitals found."

core/static/clients/patient-access/js/client-patient-access.js
```diff
@@ -0,0 +1,333 @@
...
+  // The token must carry patient context (the launch/patient scope). Without it
+  // we cannot attribute or scope the data, so stop with a clear message.
+  var epicPatientId = client.patient && client.patient.id;
+  if (!epicPatientId) {
+    out.textContent += "\nError: no patient context from Patient Access (missing launch/patient scope)";
+    return;
+  }
+  out.textContent += "\nEHR patient id: " + epicPatientId;
+
+  var idOk = await paSavePatientIdentifier(jheToken, config.iss, epicPatientId);
```
surfdoc: **Provenance uses the wrong iss.** We stamp the identifier system (here) and the FhirSource base URL (line 300) with `config.iss` — the seeded default — but the patient authorized against the hospital they picked (`row.fhirBaseUrl`). For any hospital other than the seeded default, the records get labeled with the wrong iss, which defeats the multi-hospital point of the picker and will misdirect future re-syncs.

After `FHIR.oauth2.ready()` the authorized server is available as `client.state.serverUrl` — suggest using that for both the identifier system and the FhirSource `fhir_base_url` instead of `config.iss`. (Invisible today only because the sandbox has a single iss.)

surfdoc (changes requested):

core/views/patient_access.py
```diff
@@ -0,0 +1,118 @@
...
+    patient = request.user.get_patient()
+    if patient is None:
+        return Response({"error": "Authenticated user is not a patient"}, status=400)
+
+    system = request.data.get("system")
+    value = request.data.get("value")
+    if not system or not value:
+        return Response({"error": "system and value are required"}, status=400)
+
+    PatientIdentifier.objects.get_or_create(system=system, value=value, defaults={"patient": patient})
```
surfdoc: Minor: `(system, value)` is globally unique, so if that pair is already bound to a *different* patient, `get_or_create` returns the existing row untouched and we still respond 200 with the same `{system, value}` — the caller thinks it was attached to them when it wasn't. Worth checking `created` (or `identifier.patient_id == patient.id`) and returning 409 on conflict rather than a silent success. Edge case, not a blocker.

fhir-r4-import-patches.md
```diff
@@ -0,0 +1,399 @@
+# FHIR R4 → R5 Import
```
surfdoc: Nit: this looks like working notes committed at the repo root, whereas the PR says the docs live in `software-documentation/` (`jhe/patient-access-integration.md`). Can we move it there or drop it so we don't accumulate stray top-level docs?

PR #676, opened 2026-07-26
feat(fhir): CapabilityStatement at /FHIR/R5/metadata + SMART well-known discovery

### LINKED ISSUE #615: FHIR Capability Statement
### Effort estimate (hours)

1

### Description

Read fhir_config.json and DOT meta to render a minimal FHIR capability statement

surfdoc:
> **Design + build manifest: #710** — how this was built, the interactive decisions behind it, and how the statement is kept accurate to the API. Worth reading before the diff.

## Summary

Implements #615: an R5 CapabilityStatement rendered from `fhir_config.json`
(so it cannot drift from actual routing/search behavior) plus the OAuth
authorize/token endpoints, served publicly at `GET /FHIR/R5/metadata` per the
FHIR capabilities interaction. Also serves
`GET /FHIR/R5/.well-known/smart-configuration` (SMART App Launch discovery,
what public PKCE clients fetch), with CORS on both discovery documents.

Follows the US Core server CapabilityStatement conventions as they apply here —
expectation extensions (everything declared is SHALL, since the config is an
allow-list), SMART-on-FHIR security service + oauth-uris extension, security
documentation — without claiming US Core conformance (US Core targets R4; JHE
is R5 and does not serve vread/history, Provenance `_revinclude`, or US Core
profiles).

**Dual-store honesty:** per-resource entries carry the union of the mapped and
aux stores' interactions and search params, with `documentation` naming the
caveats: mapped-store rows don't support update/delete (the view 405s them, so
those interactions are declared as imported-record-only), imported-record
creates need a FhirSource (`X-JHE-FHIR-Source-ID` or `meta.source`), and
aux-only search params note that the default JHE-native search ignores them.
`_id`/`_lastUpdated`/`_source`, aux-store `patch` (+ `patchFormat`), and the
OMH `valueAttachment` encoding on Observation are declared. The FHIR views now
also accept `Accept: application/fhir+json` (previously 406).

## Testing

- `tests/backend/test_capability_statement.py` (14 tests): unauthenticated 200
  **and** 200 with an invalid bearer, R5 validation via `fhir.resources`,
  fhir+json negotiation, CORS + Cache-Control headers, zero-DB-query build,
  X-Forwarded-Proto https URI rendering, store-caveat documentation,
  smart-configuration shape (snake_case keys preserved past the camel-case
  renderer).
- Full suite: 978 passed, 28 skipped (contract tests excluded as in CI).

## Notes

- `searchParam.definition` canonicals omitted in v1 (optional in R5).
- The smart-configuration `capabilities` list is minimal-honest
  (`launch-standalone`, `client-public`, `client-confidential-symmetric`);
  happy to extend it if JHE wants to claim more.

Closes #615


minrk (approved):
This will help me with some updates I'm working on in the client. I'm happy to merge, but it's unclear if there's any other reviews/discussions outstanding.

s1monj:
There's a table of how the values are derived here: https://github.com/jupyterhealth/jupyterhealth-exchange/issues/727#issuecomment-5343157944

From that, the `/o/` part of the auth paths should not be hardcoded as they're vars in `urls.py`. Added commits to resolve the path from config and as part of that, tidied up the auth support that was sprawled across 5 files.

Issue #391, opened 2026-05-14
Implement RFC #334 - Enhance auth for patient links with opaque secret


s1monj:
- Implementation of https://github.com/jupyterhealth/jupyterhealth-exchange/issues/334
- Testing with patient M power and CareX on UCSF
- Document

s1monj:
Tested and verified with CareX and pMp

PR #677, opened 2026-07-30
docs: RFC 0001 — CapabilityStatement source of truth + provenance for #676

surfdoc:
Companion to #676 (and #615), and a pilot of the process @s1monj proposed: ship the decision log and the design questions alongside the implementation.

Two parts:

1. **Provenance for #676** — the condensed decision log: what was asked, which choices were made by a human at each fork (US Core alignment without conformance claims, public access, smart-configuration, OMH encoding as documentation), and why the code looks the way it does.
2. **The design question** — a gap table of every fact the CapabilityStatement needs that `fhir_config.json` cannot currently express (each is a hardcoded constant in `capability.py` today), with a recommendation per row on whether it becomes config (mapped `__interaction` honesty, optional `fhir_type`, `fhir_version_number`, deriving date-sort from `__sortDate`) or stays code (view-level common params, aux PATCH invariant). On agreement, a follow-up reworks #676's builder to the extended config and deletes the constants — the existing capability tests pin behavior through the refactor.

Happy to iterate on the doc shape itself — that's part of the pilot.

### DIFF
--- a/docs/rfcs/0001-capability-statement-source-of-truth.md
+++ b/docs/rfcs/0001-capability-statement-source-of-truth.md
@@ -1,0 +1,92 @@
+# RFC 0001: CapabilityStatement source of truth (fhir_config vs. code) + AI-work provenance
+
+- **Status:** Discussion
+- **Companion PR:** #676 (CapabilityStatement + smart-configuration) · **Origin issue:** #615
+- **Also piloting:** the "share the prompt + decisions with the PR" process proposed by @s1monj
+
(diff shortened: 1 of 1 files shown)


surfdoc:
Closing — the `docs/rfcs/` approach isn't how the RFC flow is set up here. RFCs go through `.github/ISSUE_TEMPLATE/02_rfc.yml` as issues (#334 → #391), not as checked-in docs. No separate design-doc PR.

PR #679, opened 2026-07-31
chore(deps): unblock security updates — django-saml2-auth-community + uv migration (RFC 0002)

surfdoc:
## What

Unwedges the root app's dependency security pipeline, which was structurally unable to apply any pyjwt or cryptography security update:

- **Swap** `grafana-django-saml2-auth` → **`django-saml2-auth-community>=3.22.0`** (Grafana archived their repo; the community package is the maintained continuation, same import paths — zero code changes).
- **Migrate the root app pipenv → uv**, matching `/mcp_server`. Dockerfile, backend CI, and dependabot config updated.
- **Declare a one-line resolver override** (`pyopenssl>=26.2`) lifting pysaml2's stale cap — the reason uv is required: pipenv/pip/Poetry cannot express it, and our `Pipfile.lock` had drifted silently unsound working around it. Main now demonstrates the failure mode this fixes: the #692 relock shipped the cryptography 48→43 downgrade (two HIGH advisories) plus both saml distributions locked at once; this PR supersedes both.

## Result (honest resolution — the lock is consistent again)

| package | before | after |
|---|---|---|
| pyjwt | 2.12.1 (5 open advisories, 1 HIGH) | **2.13.0** |
| cryptography | 48.0.1 (unsound; 43.0.3 on main since #692) | **50.0.0** |
| pyopenssl | 24.2.1 (CVE-2026-27459) | **26.4.0** |
| django | 5.2.15 | **5.2.16** (tracks main) |

The recurring red `security_update_not_possible` dependabot job dies with this, along with the deploy-gate flakiness it caused. Dependabot #333, #695 and #696 become obsolete after merge.

## Verification

Full backend suite: **977 passed** on the refreshed lock. Import smoke across the override pairings verified; reproduction script available (RFC §7).

PR #680, opened 2026-07-31
fix(fhir-import): keep repeating primitives as lists; surface per-record import errors

surfdoc:
> **Design + build manifest: #712** — what surfaced these fixes (live Epic-sandbox testing), the debugging, and the deferred `Condition.clinicalStatus` policy call. Worth reading before the diff.

Follow-up to #671, from live end-to-end testing of the patient-access sync against the Epic sandbox (`jhe.fly.dev`, Epic test patient): **Conditions: 22 saved (45 failed), Allergies: 0 saved (3 failed)** with no visible reason.

## Fix 1: cross-version engine collapsed repeating primitives (all allergy imports failed)

`_run_rule` iterates a repeating source element one item at a time, but the `copy`/`translate` targets assigned each item to the target slot — so `AllergyIntolerance.category ["medication", "food"]` became the scalar `"food"`, and every allergy then failed R5 validation with `category: value is not a valid list`. (Epic emits a category on every allergy, hence 3/3 failed.)

Both transforms now append when the target element repeats, using the same `is_list` dispatch `_apply_create` already used. Regression test: `test_engine_copy_transform_keeps_repeating_primitive_list`.

## Fix 2: the callback UI discarded the per-record error the endpoint already returns

The import endpoint puts the create-path error in each entry's `response.outcome` (OperationOutcome), but the client only counted failures — making the 45 Condition failures undiagnosable without checking out the branch and replaying the conversion. `paWriteResource` now returns `{ok, reason}` and the callback page lists each distinct failure reason with a count, e.g.:

```
Fetching Conditions from Patient Access...
  saved 22 record(s)
  45 record(s) could not be saved:
    - Invalid FHIR Condition: clinicalStatus — field required (x45)
```

## Explicitly deferred (needs a policy call, not a code fix)

The 45 Condition failures themselves are **R5 making `Condition.clinicalStatus` mandatory** (optional in R4 — Epic returns many encounter-diagnosis/health-concern Conditions without it). Options are synthesizing `condition-clinical|unknown` during import vs. continuing to reject; with this PR the UI at least says why. Left for a follow-up issue.

## Verification
- Backend: 978 passed (incl. 31 in `test_fhir_r4_import.py`)
- Frontend: 48 passed (jest), incl. new coverage for reason extraction/aggregation
- Manual repro of the exact sandbox failure shapes through `transform_to_r5` + `validate_fhir_resource`: category list now survives and validates

## Staff-review hardening pass (second commit)
Three-lens agent review applied: `_assign` now promotes scalar residue under a repeating key (with direct unit coverage of both dispatch semantics); a backend test pins the OperationOutcome error contract the client parses (severity `error` + diagnostics string); the reasons accumulator uses a null prototype; the error-path result shape gains `reasons: {}`; and the callback caps rendered failure reasons at the 5 most frequent. Backend 33/33 in the import file, jest green.

### DIFF
--- a/core/fhir/cross_version.py
+++ b/core/fhir/cross_version.py
@@ -146,7 +146,7 @@ class _Engine:
         if transform == "copy":
             value = self._param_value(target["parameter"][0], svars, tvars)
             if element is not None and value is not None:
-                context.value[element] = value
+                self._assign(context, element, value)
             return
 
         if transform == "translate":
@@ -154,7 +154,7 @@ class _Engine:
             code = self._param_value(params[0], svars, tvars)
             url = params[1].get("valueString") if len(params) > 1 else None
             if element is not None and code is not None:
-                context.value[element] = self.maps.translate(url, code)
+                self._assign(context, element, self.maps.translate(url, code))
             return
 
         if transform in (None, "create") and element is not None:
@@ -164,6 +164,17 @@ class _Engine:
             return
 
         # ``evaluate`` (Subscription only) and any other transform are unsupported -> skipped.
+
+    @staticmethod
+    def _assign(context, element, value):
+        # A rule over a repeating source runs once per item (see _run_rule), so a repeating
+        # target must append -- plain assignment keeps only the last item and mistypes the
+        # element as a scalar (e.g. R4 AllergyIntolerance.category ["medication", "food"]
+        # collapsed to "food", failing R5 validation).
+        if is_list(context.model, element) and not isinstance(value, list):
+            context.value.setdefault(element, []).append(value)
+        else:
+            context.value[element] = value
 
     def _apply_create(self, target, svars, tvars, context, element):
         # The variable a ``create`` binds carries the *source* element into the target under the
--- a/core/fhir/cross_version.py
+++ b/core/fhir/cross_version.py
@@ -144,9 +144,18 @@ class _Engine:
         element = target.get("element")
 
         if transform == "copy":
-            value = self._param_value(target["parameter"][0], svars, tvars)
+            param = target["parameter"][0]
+            value = self._param_value(param, svars, tvars)
             if element is not None and value is not None:
-                self._assign(context, element, value)
+                source_var = svars.get(param["valueId"]) if "valueId" in param else None
+                json_key, child_model = _target_key(context.model, element, source_var)
+                # A bare code copied into a CodeableConcept-typed element (e.g. the
+                # DocumentReference4to5 map's attester.mode = "professional") is wrapped
+                # rather than assigned raw, which would fail R5 validation. No coding
+                # system is available from the map, so the code lands in ``text``.
+                if isinstance(value, str) and getattr(child_model, "__name__", None) == "CodeableConcept":
+                    value = {"text": value}
+                self._assign(context, json_key, value)
             return
 
         if transform == "translate":
--- a/core/fhir/cross_version.py
+++ b/core/fhir/cross_version.py
@@ -172,7 +172,11 @@ class _Engine:
         # element as a scalar (e.g. R4 AllergyIntolerance.category ["medication", "food"]
         # collapsed to "food", failing R5 validation).
         if is_list(context.model, element) and not isinstance(value, list):
-            context.value.setdefault(element, []).append(value)
+            existing = context.value.setdefault(element, [])
+            if isinstance(existing, list):
+                existing.append(value)
+            else:
+                context.value[element] = [existing, value]
         else:
             context.value[element] = value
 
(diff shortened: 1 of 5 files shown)


uv.lock
```diff
@@ -0,0 +1,3 @@
+version = 1
+revision = 3
+requires-python = ">=3.12"
```
travis-sauer-oltech: uv.lock was already merged into main. The python version on main is 3.11. We had to revert to 3.11 during operational support for UCSF. You can safely remove this.

core/static/clients/patient-access/js/client-patient-access.js
```diff
@@ -129,11 +142,12 @@ async function paWriteResource(jheToken, sourceId, resourceType, resource) {
...
     body: JSON.stringify(resource),
   });
-  if (!response.ok) return false;
+  if (!response.ok) return { ok: false, reason: "HTTP " + response.status };
   var bundle = await response.json();
   var entry = bundle && bundle.entry && bundle.entry[0];
   var status = entry && entry.response && entry.response.status;
-  return typeof status === "string" && status.charAt(0) === "2";
+  var ok = typeof status === "string" && status.charAt(0) === "2";
+  return { ok: ok, reason: ok ? null : paEntryFailureReason(entry) || status || "unknown error" };
```
travis-sauer-oltech, line 159: #681 adds 15 new scopes, and this line removes the response body, so a scope failure would just read as a 403. Can we include the body text?

core/fhir/cross_version.py
```diff
@@ -165,6 +165,21 @@ def _apply_target(self, target, svars, tvars):
 
         # ``evaluate`` (Subscription only) and any other transform are unsupported -> skipped.
 
+    @staticmethod
+    def _assign(context, element, value):
```
travis-sauer-oltech, line 178: This fixes AllergyIntolerance. Only `copy` has a map-level test though. Can we add one for translate onto a repeating element?

core/static/clients/patient-access/js/client-patient-access.js
```diff
@@ -324,6 +345,20 @@ async function finishPatientAccessConnect(out, config) {
...
+      out.textContent += "\n  " + result.failed + " record(s) could not be saved:";
+      // Validation messages can embed record values, making every reason distinct — cap the
+      // list at the 5 most frequent so one bad type cannot flood the page.
+      var reasonList = Object.keys(result.reasons).sort(function (a, b) {
+        return result.reasons[b] - result.reasons[a];
+      });
+      reasonList.slice(0, 5).forEach(function (reason) {
+        out.textContent += "\n    - " + reason + " (x" + result.reasons[reason] + ")";
+      });
+      if (reasonList.length > 5) {
```
travis-sauer-oltech, line 370: Can we console.log the full reasons object so a console capture keeps everything?

surfdoc:
Rebased onto current main (both PRs had gone CONFLICTING after the #679 uv migration) and added fixes for the DocumentReference import failures the review HAR captures surfaced (5/5 clinical notes failing):

- **copy into a choice element now flattens the key** (content.profile `value` → `valueCoding`) the same way create targets already did;
- **a bare code copied onto a CodeableConcept element is wrapped** (`attester.mode = "professional"` → `{text: ...}`);
- **first use of the local map-patch mechanism** (`data/fhir/fhir-cross-version-patches/`) merges the official map's two attester rules, which each created their own attester and left a mode-less one that R5 rejects.

Regression test covers all three via a real R4 clinical-note shape. 983 backend tests pass.

travis-sauer-oltech (changes requested):

uv.lock
surfdoc: Resolved by the rebase itself — after rebasing onto post-#679 main, uv.lock is no longer part of this PR's diff at all (nothing to remove).

core/static/clients/patient-access/js/client-patient-access.js
surfdoc: Fixed in 79988ae — the response body (first 300 chars) is now kept in the failure reason, so a scope rejection reads as `HTTP 403: {"detail":"..."}` instead of a bare 403. Jest case added.

core/fhir/cross_version.py
surfdoc: Added in 79988ae. No pack map translates onto a repeating element today, so the new test drives the engine with a minimal synthetic StructureMap (translate over AllergyIntolerance.category via a stub concept map) and asserts both codes survive as a list.

core/static/clients/patient-access/js/client-patient-access.js
surfdoc: Fixed in 79988ae — the full `reasons` map is `console.error`'d before the capped on-screen list, so a console capture keeps every distinct failure.

surfdoc:
Review round addressed in 79988ae — all four comments: uv.lock left the diff with the rebase; transport failures keep the response body; the full reasons map is console-logged; map-level translate-onto-repeating engine test added (synthetic map — no pack map exercises that path). Backend 984 + frontend 13 all green.

travis-sauer-oltech (approved):
@surfdoc Re-review looks good. Approved.

PR #681, opened 2026-07-31
feat(patient-access): pull all aux-supported clinical data; Observation views in FHIR browser

surfdoc:
> **Design + build manifest: #713** — the scope + Observation-view decisions, the rejected alternatives, and what got dropped (Coverage) or deferred (upsert). Worth reading before the diff.

**Stacked on #680** (shares the import-fix commits — merge #680 first, then this lands clean).

## 1. Pull everything JHE can handle today

`PATIENT_ACCESS_PULLS` grows from the 5-type demo subset to **every patient-compartment clinical type that has both an R4→R5 StructureMap and an `aux_resources` entry in `fhir_config.json`**: Conditions, Medications + Dispenses, Allergies, Immunizations, Procedures, Labs, **Vital Signs**, DiagnosticReports, DocumentReferences, Encounters, CarePlans, CareTeams, Goals, FamilyMemberHistory, ServiceRequests, Specimens, Coverage, Devices, QuestionnaireResponses. The Epic app registration (55446) already lists all of these APIs, so the only gating was our requested scopes — the seeded client config now carries one `patient/<Type>.read` per pulled type.

Not pulled (deliberately): reference/meta types (Practitioner, PractitionerRole, Location, Organization, Provenance, Medication, Binary) — they arrive by reference from the resources that cite them. DocumentReference note *content* (Binary attachments) is not fetched — the resource metadata imports fine.

Per-type failures stay isolated and, thanks to #680, now report their reasons — expect the first sandbox run to shake out R4→R5 edge cases the way allergies did.

## 2. Observation views in the FHIR Resources browser (fixes "labs don't display")

Root cause of the invisible labs: a FHIR search hits **exactly one store**, the browser never sent `_source`, and Observation is the only type living in two stores — so every browser query landed on the mapped OMH model while all 301 imported labs sat in the aux store.

Per discussion, no separate source selector: the resource dropdown now splits Observation into
- **Observation - Device Data** — the JHE-native OMH store (CGM/wearables; current behavior)
- **Observation - Labs** — aux store, `_source:below` + `category=laboratory`
- **Observation - Vital Signs** — aux store, `_source:below` + `category=vital-signs`

Both category filters use the existing aux token search (`category.coding`, bare-code matching — verified in `core/fhir/search.py`); zero server changes needed. A profile-restored plain "Observation" maps to the Device Data view.

## Deploy follow-ups (I'll handle after merge)
1. The deployed `JheClient.aux_data` scope string predates this change (seed writes it only on creation) — needs a one-time row update on fly.
2. Re-run a sandbox connect to verify the new pulls end-to-end and triage any per-type conversion failures the reasons surface.

## Verification
- Backend: 978 passed · Frontend: 48 passed
- Category token matching semantics confirmed against `core/fhir/search.py` (`system|code` and bare `code`)

## Alternatives considered

| Alternative | Why not |
|---|---|
| Separate source selector (JHE-native / imported dropdown) | Product decision — two coupled dropdowns for one mental model; named views read better. The `_source` API remains for power users. |
| Server-side cross-store union for Observation | #671 deliberately made search single-store; a union breaks pagination/count semantics and mixes provenance models. Views keep the invariant. |
| Pull everything Epic offers (Binary, EOB, Media, …) | No aux config and/or no StructureMap → the import would 4xx on every record. The line is "what we can handle today." |
| Deriving the pull list from the CapabilityStatement | Premature — the pull list also encodes Epic query constraints (e.g. Observation needs a category filter) a capability statement doesn't express. Revisit once #676's config-vs-code question settles. |

## Follow-ups

1. Sandbox end-to-end run of the expanded pull; triage surfaced conversion failures (now labeled per-record).
2. Condition `clinicalStatus` policy (deferred from #680).
3. DocumentReference Binary content pipeline; reference-resolution for cited Practitioner/Location/Medication.
4. Additional Observation category views (social-history etc.) — always in pull+view pairs (views are hardcoded client-side; a new pull without its view would be invisible again).
5. JHE patch maps (`fhir-cross-version-patches/`): Coverage `kind`/`insurer` (unblocks pulling Coverage), `Encounter.reasonCode → reason.value`.

**Merge coordination with #697:** both PRs add `core` migrations off the `0042` parent (this PR: 0043/0044; #697: 0045). Whichever merges second flips its migration's dependency to the other's leaf — one line, no rename.

## Staff-review pass (updates the description above)
- **Coverage is no longer pulled** (and its scope removed): R5's mandatory `Coverage.kind` has no StructureMap rule, so every record failed validation — verified empirically. Excluded until a patch map exists (follow-up 5).
- **Epic-required search filters added**: `CarePlan?category=assess-plan`, `CareTeam?status=active`, `DocumentReference?category=clinical-note`.
- **Observation pulls dedupe within a run** (dual-categorized records — e.g. POC glucose in both laboratory and vital-signs — import once).
- **Views generalized beyond Observation**: Patient and Device (the other mapped types the sync writes) get JHE-native / Imported-EHR dropdown views — imported Patient rows were invisible since #671, and the Device pull would otherwise have recreated the bug.
- **Status-less R4 Conditions now import with `clinicalStatus = unknown`** plus a per-record OperationOutcome warning (product decision; R4 import path only — the native R5 API is unchanged). The 45 previously-rejected sandbox Conditions now land.
- **Invariant tests**: parametrized convert+validate over every pulled type (the test that would have caught Coverage), scope↔pull set equality, and enrichment behavior. Backend 1003 passed, frontend 48 passed.


uv.lock
```diff
@@ -0,0 +1,3 @@
+version = 1
+revision = 3
+requires-python = ">=3.12"
```
travis-sauer-oltech: Same request as here: https://github.com/jupyterhealth/jupyterhealth-exchange/pull/680/changes#r3719939846.

You can safely remove this entire uv.lock file.

core/views/fhir_import.py
```diff
@@ -88,8 +88,9 @@ def _process_resource(self, resource_type, body, already_camel=False):
                     resource_type,
                     ", ".join(dropped),
                 )
+            enriched = _enrich_r5(resource_type, r5)
```
travis-sauer-oltech, line 120: if `_create` raises below, `_error_entry` is called without enriched, and the warning is lost.

We should pass it through.

core/views/fhir_import.py
```diff
@@ -102,9 +103,35 @@ def _convert(self, resource_type, camel_body):
...
-def _success_entry(created, resource_type, dropped):
+def _enrich_r5(resource_type, r5):
+    """Default R5-mandatory fields that are optional in R4 and absent at the source.
+
+    Returns the warning issues describing what was defaulted (empty when nothing was).
+    Today: R4 Condition.clinicalStatus is 0..1 but R5 requires it; Epic emits many
+    encounter-diagnosis/health-concern Conditions without one. Rather than reject the
+    record, default to the value-set's own escape hatch 'unknown' and say so per record.
+    """
+    if resource_type == "Condition" and not r5.get("clinicalStatus"):
```
travis-sauer-oltech, line 143: This warning never surfaces. `paPullResourceType` only prints reasons for failed records. 45 conditions get a status written in with nothing on screen. The RFC says this is visible, not silent.

Let's surface it or adjust the claim.

core/management/commands/seed.py
```diff
@@ -265,7 +265,20 @@ def _generate_client_id(length=40):
                     # jhe.fly.dev /clients/patient-access/callback redirect URIs registered.
                     # (Same Epic sandbox app/client id used in the Phase 1 POC.)
                     "client_id": "77849e74-8e2a-4c2f-826c-bdbef6da3357",
-                    "scopes": "openid profile launch/patient patient/Patient.read patient/Observation.read patient/Condition.read patient/MedicationRequest.read patient/AllergyIntolerance.read",
+                    # One read scope per PATIENT_ACCESS_PULLS type (client-patient-access.js);
+                    # the Epic app registration already covers all of these APIs.
+                    "scopes": (
```
travis-sauer-oltech, line 270: Seed only writes aux_data on creation. Merging this includes a client asking for 15 scopes the deployed row doesn't have.

Can we change this to be a migration instead of a post-merge step?

core/static/clients/patient-access/js/client-patient-access.js
```diff
@@ -129,26 +142,51 @@ async function paWriteResource(jheToken, sourceId, resourceType, resource) {
...
 }
 
-// USCDI resources pulled for the demo phenotype. `single` reads one instance (Patient),
-// the rest are patient-scoped searches. Order is display order.
+// Every patient-compartment clinical type JHE can ingest today: each has an R4->R5
+// StructureMap and an aux_resources entry in fhir_config.json (reference/meta types like
+// Practitioner, Location and Provenance are resolved from the resources that cite them, not
+// pulled). `single` reads one instance (Patient), the rest are patient-scoped searches.
+// Order is display order. Failures are isolated per type and reported with reasons.
 var PATIENT_ACCESS_PULLS = [
```
travis-sauer-oltech, line 222: 20 types, one async POST per record. If there are hundreds of labs this equates to thousands of round trips. `fhir-import/R4` already takes a bundle.

Can we adjust this to post to as a bundle?

core/static/clients/patient-access/js/client-patient-access.js
```diff
@@ -159,19 +197,30 @@ async function paPullResourceType(client, jheToken, sourceId, pull, iss) {
...
   var failed = 0;
+  // Distinct failure text -> count, so 45 identical errors read as one line. Null prototype:
+  // a diagnostics string like "__proto__" or "constructor" must count as a plain key.
+  var reasons = Object.create(null);
   for (var i = 0; i < resources.length; i++) {
     var resource = resources[i];
     if (!resource || resource.resourceType !== pull.type) continue;
+    if (seenIds && resource.id) {
+      if (seenIds.has(resource.id)) continue;
+      seenIds.add(resource.id);
```
travis-sauer-oltech: Added before the write. A record that fails in a Labs pull gets skipped by Vital Signs instead of retried.

Can we move this to after a successful write?

core/static/clients/patient-access/js/client-patient-access.js
```diff
@@ -129,26 +142,51 @@ async function paWriteResource(jheToken, sourceId, resourceType, resource) {
...
+// Order is display order. Failures are isolated per type and reported with reasons.
 var PATIENT_ACCESS_PULLS = [
   { label: "Demographics", type: "Patient", query: "Patient", single: true },
   { label: "Conditions", type: "Condition", query: "Condition" },
   { label: "Medications", type: "MedicationRequest", query: "MedicationRequest" },
+  { label: "Medication Dispenses", type: "MedicationDispense", query: "MedicationDispense" },
   { label: "Allergies", type: "AllergyIntolerance", query: "AllergyIntolerance" },
+  { label: "Immunizations", type: "Immunization", query: "Immunization" },
+  { label: "Procedures", type: "Procedure", query: "Procedure" },
+  // Epic requires a category (or code) filter on Observation searches, so each pulled
```
travis-sauer-oltech, line 230: Re-running Connect still duplicates everything.

Can the upsert land before we widen the pull list?

core/static/clients/jhe-admin/js/client-jhe-admin.js
```diff
@@ -1889,12 +1889,48 @@ async function renderFhir(queryParams) {
...
+  };
+  // viewName -> {path, query}; and mapped-type name -> its first view (for restores of the
+  // plain path the server remembers, and old ?resource= bookmarks).
+  const viewsByName = {};
+  for (const [path, views] of Object.entries(RESOURCE_VIEWS)) {
+    for (const [name, query] of Object.entries(views)) {
+      viewsByName[name] = { path, query };
+    }
+  }
+  if (RESOURCE_VIEWS[queryParams.resource]) {
```
travis-sauer-oltech, line 1950: The server saves the URL path (`Observation`), not the view name.

See `core/views/fhir.py` line 472.

Someone on "Observation - Labs" who navigates away and back lands on Device Data and sees zero labs.

core/static/clients/jhe-admin/js/client-jhe-admin.js
```diff
@@ -1889,12 +1889,48 @@ async function renderFhir(queryParams) {
...
 
-  // The Resource select lists every supported FHIR resource type (from server settings).
+  // Six types are "mapped" (served from native Django models when no _source is sent) and
+  // also accept imported EHR rows into the aux store; a search hits exactly ONE store (see
+  // core/views/fhir.py). For the mapped types the sync pulls (Observation, Patient, Device),
+  // the picker splits the entry into per-store views instead of exposing a source selector —
+  // otherwise imported rows are unreachable from the browser.
+  // Mirrors JHE_FHIR_SOURCE_BASE (core/models/fhir_aux_resource.py); the server rstrips the slash.
+  const IMPORTED_SOURCE_PREFIX = "https://jupyterhealth.org/fhir/fhir-source/";
+  const RESOURCE_VIEWS = {
```
travis-sauer-oltech, line 1902: This covers 3 of the 6 mapped types. Group, Organization, and Practitioner are mapped and have aux entries. Imported rows of those stay invisible.

Can we cover this or make a line note?

tests/backend/test_fhir_r4_import.py
```diff
@@ -483,3 +541,129 @@ def test_import_bundle_reports_per_entry_error(api_client, patient, fhir_source)
...
+# ---------------------------------------------------------------------------
+
+_REPO_ROOT = pathlib.Path(__file__).resolve().parents[2]
+_SEED_PY_PATH = _REPO_ROOT / "core" / "management" / "commands" / "seed.py"
+_CLIENT_JS = _REPO_ROOT / "core" / "static" / "clients" / "patient-access" / "js" / "client-patient-access.js"
+
+
+def _pulled_types():
+    text = _CLIENT_JS.read_text()
+    block = re.search(r"var PATIENT_ACCESS_PULLS = \[(.*?)\n\];", text, re.DOTALL).group(1)
```
travis-sauer-oltech: This is unguarded. For example; AttributeError on None would surface instead of a readable failure.

tests/backend/test_fhir_r4_import.py
```diff
@@ -483,3 +541,129 @@ def test_import_bundle_reports_per_entry_error(api_client, patient, fhir_source)
...
+    from core.views.fhir_import import _enrich_r5
+
+    r5 = transform_to_r5(resource_type, _MINIMAL_PULL_R4[resource_type])
+    _enrich_r5(resource_type, r5)
+    validate_fhir_resource(resource_type, r5)
+
+
+def test_pulled_types_match_seeded_scopes():
+    seed_text = _SEED_PY_PATH.read_text()
+    scoped = set(re.findall(r"patient/(\w+)\.read", seed_text))
```
travis-sauer-oltech: This scans all of seed.py. Any future client with patient scopes breaks the test incorrectly.

Can we scope this to the patient-access block?

core/static/clients/jhe-admin/js/client-jhe-admin.js
```diff
@@ -1889,12 +1889,48 @@ async function renderFhir(queryParams) {
...
+    Observation: {
+      "Observation - Device Data": {},
+      "Observation - Labs": { "_source:below": IMPORTED_SOURCE_PREFIX, category: "laboratory" },
```
travis-sauer-oltech, line 1905: Can we add a test? import a laboratory Observation, then assert `?_sourece:below=<base>/&category=laboratory` returns it and that the plain search does not?

surfdoc:
Rebased onto the updated #680 (both PRs had gone CONFLICTING after #679) plus pull-list fixes from the live-capture findings:

- **FamilyMemberHistory and Specimen are fully removed** — pull, seeded scope, and their `fhir_config.json` aux_resources entries (so they also leave the FHIR browser dropdown and the import surface). Verified live: Epic 403s FamilyMemberHistory on the patient-access API and 400s `Specimen?patient=` (no patient-level Specimen search); patient-access EHR APIs do not serve either type. Re-adding one is a single config entry if a source ever appears.
- **Device now queries with an explicit `?patient=`** via plain `client.request` — fhir-client's `patient.request` has no compartment param mapping for it and threw client-side before any request was made. Epic accepts the query (valid empty bundle for the sandbox patients).

1003 backend tests pass. Fresh HAR captures of the full flow (Epic sandbox patients, zero fetch failures, Documents importing 5/5 via the #680 engine fixes) accompany this review — ping @PatrickCarter for the files.

travis-sauer-oltech (changes requested):

uv.lock
surfdoc: Same as #680 — the rebase onto post-#679 main dropped uv.lock from this diff entirely.

core/views/fhir_import.py
surfdoc: Fixed in 92de366 — `enriched` is initialized before the try and passed through to `_error_entry`, so the defaulted-clinicalStatus warning survives a failed create (which is exactly when it narrows the diagnosis). Regression test added.

core/views/fhir_import.py
surfdoc: Fixed in 92de366 — surfaced, not softened: successful entries' warnings now flow back through `paWriteResource`/`paWriteBundle` and the connect log renders "some saved record(s) were adjusted during import" with the same capped-list + full `console.warn` treatment as failures. The RFC's 'visible, not silent' claim now holds.

core/management/commands/seed.py
surfdoc: Done in 92de366 — migration 0043 updates existing Patient Access rows in place. It touches only the `scopes` key (a deployed `client_id` or any operator-set aux_data survives), and a test guards the migration/seed.py/PATIENT_ACCESS_PULLS lockstep so the three can't silently diverge.

core/static/clients/patient-access/js/client-patient-access.js
surfdoc: Done in 92de366 — the pull now posts chunked batch Bundles (100 records/POST) to `fhir-import/R4`'s bundle route via a new `paWriteBundle`, mapping the order-aligned batch-response entries back to per-record results. A few hundred labs is now a handful of round trips.

core/static/clients/patient-access/js/client-patient-access.js
surfdoc: Fixed in 92de366 — a record is marked seen only after a successful write, so a failure in the Labs pull is retried when Vital Signs returns the same record. Test covers fail → retry → dedupe.

core/static/clients/jhe-admin/js/client-jhe-admin.js
surfdoc: Fixed in 92de366, at the root you pointed at: `_remember_resource` now persists the view-defining search params (`category`, `_source:below`) alongside the path — pagination/org/study context params are excluded — and the client resolves `Observation?category=laboratory&...` back to "Observation - Labs" on restore. Plain legacy values keep the old first-view fallback. Tested on both sides.

core/static/clients/jhe-admin/js/client-jhe-admin.js
surfdoc: Gap closed in ad31fe3 — every mapped type now splits into native + Imported EHR views ("Group - Studies" / "Group - Imported EHR", "Organization - JHE" / "Organization - Imported EHR", "Practitioner - JHE" / "Practitioner - Imported EHR"), so rows imported directly through /fhir-import are reachable from the browser. Backend test pins the view query for a native-mapped type.

tests/backend/test_fhir_r4_import.py
surfdoc: Fixed in 92de366 — the regex miss now fails with a readable assert ("PATIENT_ACCESS_PULLS not found — was it renamed or reformatted?") instead of an AttributeError on None.

tests/backend/test_fhir_r4_import.py
surfdoc: Fixed in 92de366 — the parse is scoped to the Patient Access client's aux_data scopes block in seed.py, so a future client with its own patient/*.read scopes can't break the invariant.

core/static/clients/jhe-admin/js/client-jhe-admin.js
surfdoc: Added in 92de366 — imports a laboratory Observation, asserts the exact Labs-view query (`_source:below=<base>/&category=laboratory`) returns it, and that the plain Observation search (mapped/device-data store) does not.

core/static/clients/patient-access/js/client-patient-access.js
surfdoc: Done in a798523 — folded into this PR after discussion (Patrick agreed it shouldn't ship separately). `create_aux_resource` now upserts on `(fhir_source, resource_type, fhir_resource_id)`: re-running Connect (or re-POSTing the same EHR record) refreshes the existing row in place, keeping its JHE UUID so stored references stay valid. Bodies with no upstream id still always create. Migration 0044 first collapses any duplicates a deployment already accumulated (keeping the most recently updated row), then adds a conditional unique constraint backing the upsert against races. Update-in-place was chosen over skip so a refreshed EHR record wins. Covered at both the aux layer (same-source upsert, cross-source separation, no-id create) and the import endpoint (re-import idempotency, stable JHE id).

surfdoc:
Review round addressed in 92de366 + ad31fe3 + a798523 (rebased onto the updated #680 — your comment threads are all preserved; a couple show as *outdated* in Files changed but remain in this Conversation tab). All 13 comments implemented, including: enrichment warnings surviving failed creates, saved-with-changes warnings surfaced in the connect log, chunked Bundle imports, dedupe marking only after successful writes, migration 0043 for the deployed client's scopes, the Labs-view restore fixed at the server (view-defining params are remembered), Imported-EHR browser views for Group/Organization/Practitioner (ad31fe3), and the upsert itself (a798523): aux rows are unique per (source, type, upstream id) with migration 0044 deduping existing deployments — re-running Connect now refreshes instead of duplicating. RFC 0003 cons/follow-ups updated to match. Backend 1013 + frontend 53 all green.

surfdoc:
Ran a multi-agent adversarial review over both PRs (six attack agents by subsystem, every finding independently refute-verified twice). It confirmed 8 real defects — most importantly that the upsert in a798523 did **not** actually deliver cross-run idempotency: the Connect flow registers a fresh FhirSource every run, so the source-scoped upsert key never matched and a reconnect still duplicated the whole chart. All 8 are fixed in 8ae5ee3:

- **FhirSource registration is now idempotent** per (patient, data source, base URL) — the root cause of cross-run duplication
- **Over-64-char Epic ids** are sanitized server-side and still key the upsert (previously the client deleted them, defeating both dedupe layers)
- **Pull isolation holds under transport failure** — a rejected fetch/truncated response is reported per record instead of freezing the page mid-connect
- **Malformed bundle entries** fail as their own entry (earlier entries are already committed; a whole-request 422 misreported what was written)
- **Bundle route honors per-entry meta.source** when no header is sent, matching its own error text
- **View restore survives an active study filter** (the ephemeral-params set now uses the post-middleware `_group` key spelling)
- **Sticky-view memory is opt-in** via an `X-JHE-Remember-View` header only the admin browser sends — MCP/API searches no longer clobber it
- **Issuer-less Device.udiCarrier** (legal R4, required in R5) is dropped per-carrier with a warning instead of rejecting the Device
- **Migration 0044 repoints ref-indexed references** at the kept row before deleting duplicates, so nothing dangles

Refuted by the verification pass (no change needed): the 2.5MB request-size concern, a Postgres-error-to-patient race, migration 0043 scope-clobbering, and a fly release-command race. #680 itself came out clean. Backend 1020 + frontend 54 all green.

travis-sauer-oltech (approved):
@surfdoc LGTM. Approved. Thank you for addressing all of the review items and running a multi-agent adversarial review

PR #694, opened 2026-08-03
fix default org roles setting

minrk:
env migration appeared to make it single characters, which didn't produce valid relationships, since `v` and `m` are not valid roles. This produced invalid relationships that do not confer the desired permissions. Incorrect validation also prevented _correct_ initial org relationships from being allowed.

regression in #298 prevented default orgs setting from working:

- correct `123:viewer` would fail validation and create a non-functional JheUser (failure should have rolled back creation of the user)
- incorrect `123:m` would pass validation and create a member with a non-functional role (should have failed validation in creating the invalid PatientOrganization relationship, but didn't)

This fixes the correct `123:viewer` to being supported.

encountered by @kirkbyl on the Berkeley demo JHE today


core/models/jhe_user.py
```diff
@@ -143,7 +143,7 @@ def save(self, *args, **kwargs):
                         if not parts:
                             raise DjangoValidationError("PRACTITIONER_DEFAULT_ORGS must be non-empty when set.")
 
-                        valid_roles = {c[0] for c in PractitionerOrganization.ROLE_CHOICES}
+                        valid_roles = set(PractitionerOrganization.ROLE_CHOICES.keys())
```
travis-sauer-oltech, line 202: `super().save()` on 117 commits the user before the atomic block on 128, so a bad `auth.default_orgs` leaves an orphan user behind. Can we wrap the whole `save()`?

core/models/jhe_user.py
```diff
@@ -143,7 +143,7 @@ def save(self, *args, **kwargs):
                         if not parts:
                             raise DjangoValidationError("PRACTITIONER_DEFAULT_ORGS must be non-empty when set.")
```
travis-sauer-oltech: These messages name `PRACTITIONER_DEFAULT_ORGS` but the value is read from `auth.default_orgs`, and that env var isn't used anywhere. Can we rename them and drop the setting?

core/management/commands/seed.py
```diff
@@ -92,7 +92,7 @@ def seed_jhe_settings(self):
             ("site.ui.title", "string", "JupyterHealth Exchange"),
             ("site.time_zone", "string", "America/Los_Angeles"),
             ("site.registration_invite_code", "string", invite_code),
-            ("auth.default_orgs", "string", ""),  # "20001:v;20002:m"
+            ("auth.default_orgs", "string", ""),  # "20001:viewer;20002:member"
```
travis-sauer-oltech, line 95: Can we add a migration file or an upgrade note? Existing installs will have the old values stored, which will now hard fail at sign-up. Additionally, the old rows never get updated with the new roles.

tests/backend/test_get_setting_migration.py
```diff
@@ -328,9 +328,8 @@ def tearDown(self):
 
     @patch(GET_SETTING_USER)
     def test_practitioner_assigned_to_default_org(self, mock_gs):
-        # ROLE_CHOICES is a dict; valid_roles = {c[0] for c in dict} gives first char of keys
-        # So valid roles are 'm', 'v' (from 'member', 'manager', 'viewer')
-        mock_gs.return_value = f"{self.org.id}:v"
+        # valid roles are keys in ROLE_CHOICES
+        mock_gs.return_value = f"{self.org.id}:viewer"
```
travis-sauer-oltech: Can we add a case asserting an invalid role raises?

tests/backend/test_get_setting_migration.py
```diff
@@ -339,7 +338,9 @@ def test_practitioner_assigned_to_default_org(self, mock_gs):
         )
         practitioner = Practitioner.objects.get(jhe_user=user)
         self.assertTrue(
-            PractitionerOrganization.objects.filter(practitioner=practitioner, organization=self.org, role="v").exists()
+            PractitionerOrganization.objects.filter(
```
travis-sauer-oltech: Can we add a two-org case with different roles so the `;` split is tested?

travis-sauer-oltech (dismissed):

core/management/commands/seed.py
minrk: I'm not sure - I don't think anyone can have the old values, since they never worked. If you followed the docs you would immediately get invalid users whenever they were created.

When this was documented and worked, it was of the form `"20001:viewer;20002:member"`. During the migration to db settings, these comments were changed to these abbreviated values that have never worked.

core/models/jhe_user.py
minrk: done.

tests/backend/test_get_setting_migration.py
minrk: more test cases covering invalid values and making sure we don't leave orphaned User entries anymore

tests/backend/test_get_setting_migration.py
minrk: added

core/models/jhe_user.py
minrk: done, and the issue is now covered in tests

core/management/commands/seed.py
minrk: This remains the only unaddressed issue, and I don't think there's a change to make, unless I misunderstand. Prior to this PR, deployments would have either:

- `2001:v` (documented-ish in a single comment, but incorrect) in which case users would be created, and get an _invalid_ org assignment, granting no permissions (equivalent to not setting the value at all)
- `2001:viewer` (correct, matching prior env config) in which case the user would be _partially_ created, get no org assignments, and first login would fail. Subsequent logins would succeed, again equivalent to not having set `auth.default_orgs`

both of which are wrong in different ways.

where after this PR, they would get

- `2001:v` - error, no user created (correct)
- `2001:viewer` - org assignments created (correct)

so I don't think we need a migration note from a never-working config value.

tests/backend/test_get_setting_migration.py
```diff
@@ -316,42 +318,86 @@ def test_email_contains_db_site_url(self, mock_gs):
...
-            identifier="ndo1",
-            user_type="practitioner",
-        )
-        practitioner = Practitioner.objects.get(jhe_user=user)
-        self.assertEqual(PractitionerOrganization.objects.filter(practitioner=practitioner).count(), 0)
+    # make sure user didn't get partially created
+    assert JheUser.objects.filter(email=email).first() is None
+
+
+def test_empty_default_orgs_skips_assignment(mock_gs):
```
minrk, line 391: btw, this shifts the default org tests to standard pytest tests with fixtures, which I think are easier to work with and simpler than unittest TestCase classes. I'd recommend most future tests being written this way. Not a big deal, but much easier to parametrize, write asserts, etc.

travis-sauer-oltech: The shift sounds fine. I know we had a contributing md that we were using to store project specifics. Do you think we should revitalize that and add this? No big deal if not.

core/management/commands/seed.py
travis-sauer-oltech: You're right it never worked, and at this stage of deployments, it won't effect anything.

travis-sauer-oltech:
@minrk LGTM. Feel free to merge.

PR #697, opened 2026-08-04
feat(auth): migrate SAML SSO to django-allauth[saml]; retire the resolver override (stacked on #679)

surfdoc:
> **Design + build manifest: #711** — how this was built, the interactive decisions behind it, and the security-sensitive design choices (existing-account handling, login-CSRF stance, xmlsec finding). Worth reading before the diff.

Continuation of #679 review: SAML joins patient login and password reset on allauth. `django-allauth[saml]` replaces `django-saml2-auth-community`, ~250 lines of vendored ACS/user-creation code are deleted, the pysaml2/pyopenssl chain disappears — **retiring the `[tool.uv]` resolver override** (#679's stated retirement trigger) — and the xmlsec1 deployment gap closes with **zero Dockerfile changes** (python3-saml signs in-process via the bundled xmlsec wheel; verified in the built image).

**Design record** (per the repo's issue-flow RFC convention, this PR description is the durable record; the earlier `docs/rfcs/0003` file is removed): operator setup — IdP SocialApp config, signing, stable-uid requirement, and back-out — lives in the docs update, jupyterhealth/software-documentation#115, sequenced to merge with this PR.

**Provisioning semantics** (@travis-sauer-oltech — the part I'd most value your read on): every IdP-asserted user is created as a practitioner (`JheUser.save()` auto-creates the profile and applies `auth.default_orgs`), `identifier` = the SAML uid, `email_is_verified` set. No asserted email → fails closed ("Sign Up Closed"). Existing same-email accounts log straight in via per-IdP email-authentication and get linked; existing **patient** accounts are rejected. The `auth.sso.valid_domains` gate applies only to first-time provisioning.

**Accepted risks**: `SOCIALACCOUNT_LOGIN_ON_GET` (drive-by *self*-login only — attacker-account substitution is blocked by allauth's session-bound `InResponseTo` check and default rejection of IdP-initiated SSO); the per-IdP `"email_authentication"`/`"verified_email"` keys are security-relevant (only for IdPs that verify mailbox ownership — note allauth's default attribute mapping also honors an IdP-asserted `email_verified` attribute); SP signing flags default off in allauth (the docs' setup guide requires the `advanced` block).

## Hardening (from a 3-way adversarial pre-review, verified against allauth 65.18 source)

- **Existing practitioners can log in via SAML**: email-authentication + auto-connect, enabled per-IdP by `"email_authentication": true` in the SocialApp JSON (the matched email must also count as verified — IdP-asserted `email_verified` or `"verified_email": true`). Without it, allauth's enumeration-safe default dead-ends every existing same-email user. The adapter backfills the verified allauth `EmailAddress` row at link time (else allauth wipes the practitioner's password) and rejects existing **patient** accounts — SAML is a practitioner entrance.
- **No login-page 500 on misconfiguration**: the SAML button renders only when the `auth.sso.saml2` flag is on *and* exactly one saml SocialApp exists; otherwise it hides.
- IdPs asserting no email fail closed (no type-any-email signup form); latent `ACCOUNT_USER_MODEL_USERNAME_FIELD` misconfiguration fixed.

## Verification

**994 backend tests** (15 SAML-specific: SP metadata through the full python3-saml/xmlsec chain, the button guards, and end-to-end ACS flows — provisioning, existing-practitioner password survival, patient rejection, domain rejection) · repo-pinned pre-commit green · Docker image build + in-container import check.

**Merge coordination with #681:** both PRs add `core` migrations off the `0042` parent (#681: 0043/0044; this PR: 0045). Whichever merges second flips its migration's dependency to the other's leaf — one line; a comment in `0045_remove_idp_metadata_url_setting.py` says exactly what to change.

cc @travis-sauer-oltech — this replaces the integration you authored; the provisioning-semantics section above is the part I'd most value your read on.


jhe/settings.py
```diff
@@ -107,6 +111,22 @@
...
+# setting above, which would bounce SSO users through email verification).
+SOCIALACCOUNT_EMAIL_VERIFICATION = "none"
+# GET-initiated login is acceptable for our own login-page redirect; skips
+# allauth's interstitial confirm page.
+SOCIALACCOUNT_LOGIN_ON_GET = True
+# Log an SSO user into an existing same-email account instead of allauth's
+# enumeration-safe dead end, and link the SocialAccount on first use. Only
+# takes effect for emails the IdP's SocialApp marks trusted via
+# "verified_email": true in its settings JSON (see docs/rfcs/0003 §4).
+SOCIALACCOUNT_EMAIL_AUTHENTICATION = True
```
travis-sauer-oltech: I think this wipes existing practitioners' passwords. lookup() -> _lookup_by_email() -> _accept_login() -> wipe_password(), which sets an unusable password if there's no verified allauth EmailAddress row. Practitioners don't have one - signup() sets the password directly, and only the patient OTP view writes EmailAddress rows. 

Can we backfill in pre_social_login?

core/templates/registration/login.html
```diff
@@ -26,7 +27,7 @@
 
 <p class="mt-3">
 {% if SAML2_ENABLED %}
-    <a class="btn btn-primary" href="{% url 'saml_signin' %}">Continue with SAML2</a>
+    <a class="btn btn-primary" href="{% provider_login_url 'saml' next='/clients/jhe-admin/organizations' %}">Continue with SAML2</a>
```
travis-sauer-oltech: Hardcoded next means SAML ignores ?next= while password login on this same page honors it. `request.GET.next|default:'/clients/jhe-admin/organizations'`?

docs/rfcs/0003-saml-on-allauth.md
```diff
@@ -0,0 +1,109 @@
...
+stack + vendored code + override forever; Simon's direction was explicit);
+dropping SAML (a product decision nobody made — allauth keeps it at near-zero
+cost); IdP config in JheSettings via a custom `get_app` (more glue for a
+setting never once populated; SocialApp admin is the native pattern).
+
+## 3. Operational runbook (first real IdP)
+
+1. Django admin → Social applications → add: provider `saml`, a name,
+   `client_id` = the URL slug (e.g. `berkeley`). IdP metadata URL, attribute
+   mapping, and optional SP cert/key go in the settings JSON
```
travis-sauer-oltech: We used to pin AUTHN_REQUESTS_SIGNED/SIGN_REQUEST to `not DEBUG`; allauth's build_saml_config defaults want_assertion_signed, want_message_signed and authn_request_signed to False. Can the runbook show the required advanced block?

core/adapters.py
```diff
@@ -19,3 +22,32 @@ class JheAccountAdapter(DefaultAccountAdapter):
...
+class JheSocialAccountAdapter(DefaultSocialAccountAdapter):
+    """SAML SSO glue: IdP-asserted users are provisioned as practitioners,
+    optionally restricted to email domains in the `auth.sso.valid_domains`
+    JheSetting (comma-separated; empty allows any). The gate applies to
+    first-time provisioning only — an already-linked SocialAccount keeps
+    logging in if the setting is tightened later; deactivate the user to
+    revoke access. Rejected signups get allauth's "Sign Up Closed" page.
+    """
+
+    def is_open_for_signup(self, request, sociallogin):
```
travis-sauer-oltech, line 64: AUTO_CONNECT skips this for existing users, patients included.  A patient in the IdP directory would land in the admin UI. 

Can we reject non-practitioners in pre_social_login?

jhe/settings.py
```diff
@@ -107,6 +111,22 @@
...
+# SAML SSO via allauth's saml provider. IdP config (metadata URL, attribute
+# map, SP certs) is a per-deployment SocialApp row in Django admin; the login
+# button is gated by the auth.sso.saml2 JheSetting. See docs/rfcs/0003.
+SOCIALACCOUNT_ADAPTER = "core.user_messages.JheSocialAccountAdapter"
+# "none": the IdP asserts the email (default falls through to the account
+# setting above, which would bounce SSO users through email verification).
+SOCIALACCOUNT_EMAIL_VERIFICATION = "none"
+# GET-initiated login is acceptable for our own login-page redirect; skips
+# allauth's interstitial confirm page.
+SOCIALACCOUNT_LOGIN_ON_GET = True
```
travis-sauer-oltech, line 127: This is login CSRF. Your RFC accepts it, but can we say why it's fine here?

jhe/settings.py
```diff
@@ -107,6 +111,22 @@
...
+# Log an SSO user into an existing same-email account instead of allauth's
```
travis-sauer-oltech: Lines 124-127; `can_authenticate_by_email` (`adapter.py` line 343) doesn't look at `verified_email`, it falls back to this global True.

An address also counts verified when the IdP asserts `email_verified` (see `provider.py` line 25). Your RFC 52-53 says the same thing, and Section 4 is xmlsec, not this.

tests/backend/test_saml_allauth.py
```diff
@@ -0,0 +1,126 @@
...
+    def test_button_hidden_when_multiple_social_apps(self):
+        SocialApp.objects.create(provider="saml", name="IdP A", client_id="a")
+        SocialApp.objects.create(provider="saml", name="IdP B", client_id="b")
+        _set_setting("auth.sso.saml2", "int", 1)
+        response = self.client.get("/accounts/login/")
+        self.assertEqual(response.status_code, 200)
+        self.assertNotContains(response, "Continue with SAML2")
+
+
+class JheSocialAccountAdapterTests(TestCase):
```
travis-sauer-oltech, line 87: Nothing covers existing-practitioner login via EMAIL_AUTHENTICATION. That's where the password wipe happens, There's no end-to-end ACS test. Signup-closed rendering and the patient case too.

docs/rfcs/0003-saml-on-allauth.md
```diff
@@ -0,0 +1,109 @@
+# RFC 0003: SAML SSO on allauth — one auth stack, no vendored ACS, no override
+
+- **Status:** Discussion
+- **Companion PR:** #697 (stacked on #679 / RFC 0002)
+- **Primary reviewers:** @s1monj (architecture), @travis-sauer-oltech (SAML behavior — author of the integration this replaces)
+- **Follows:** RFC 0002 §5's resolution; provenance per RFC 0001
```
travis-sauer-oltech: `docs/rfcs/` only has 0002 and 0003, but no 0001. Commit message references it too.

docs/rfcs/saml-on-allauth.md
```diff
@@ -0,0 +1,109 @@
+# RFC 0003: SAML SSO on allauth — one auth stack, no vendored ACS, no override
+
+- **Status:** Discussion
+- **Companion PR:** #697 (stacked on #679 / RFC 0002)
+- **Primary reviewers:** @s1monj (architecture), @travis-sauer-oltech (SAML behavior — author of the integration this replaces)
+- **Follows:** RFC 0002 §5's resolution; provenance per RFC 0001
+
+**TL;DR:** Practitioner SAML SSO moves from the vendored django_saml2_auth
+integration to allauth's built-in SAML provider (`django-allauth[saml]`) —
```
travis-sauer-oltech: Two things here. Nothing links to the RFCs (0002 either).

Can we add a list?

And` jhe/auth.md` in the software-documentation repository still documents the old setup: the `django-saml2-auth link`, `auth.sso.idp_metadata_url`, and `/sso/acs/` as the ACS URL.

We need to update the software-documentation repo to match these updates before this goes in.

docs/rfcs/saml-on-allauth.md
```diff
@@ -0,0 +1,109 @@
...
+
+Alternatives rejected: staying on the community fork (permanent second auth
+stack + vendored code + override forever; Simon's direction was explicit);
+dropping SAML (a product decision nobody made — allauth keeps it at near-zero
+cost); IdP config in JheSettings via a custom `get_app` (more glue for a
+setting never once populated; SocialApp admin is the native pattern).
+
+## 3. Operational runbook (first real IdP)
+
+1. Django admin → Social applications → add: provider `saml`, a name,
```
travis-sauer-oltech: extract_uid falls back to NameID. If one happens to be transient, it gives a new SocialAccount every login and junk in JheUser.identifier. 

Can we consider the runbook requiring a mapped uid?

core/adapters.py
```diff
@@ -19,3 +22,32 @@ class JheAccountAdapter(DefaultAccountAdapter):
         "unknown_email": _("We couldn't find an account for that email address."),
         "rate_limited": _("You're making requests too quickly. Please wait a moment and try again."),
     }
+
+
+class JheSocialAccountAdapter(DefaultSocialAccountAdapter):
```
travis-sauer-oltech, line 29: This file is for reworded error messages.

Can we create a `core/adapters.py` for both?

core/context_processors.py
```diff
@@ -28,6 +29,16 @@ def _get_oidc_client_id():
...
 
 
+def _saml2_enabled():
+    # The login button's provider_login_url resolves only when exactly one saml
+    # SocialApp exists (0 raises DoesNotExist, >1 MultipleObjectsReturned, either
+    # of which would 500 the whole login page). Hide the button instead;
+    # multi-IdP deployments need per-IdP buttons in a customized template.
+    if not get_setting("auth.sso.saml2", 0):
+        return False
+    return SocialApp.objects.filter(provider="saml").count() == 1
```
travis-sauer-oltech: Once SAML is on this is a DB query on every page render. `get_setting` and `_get_oidc_client_id` are both cached.

Can we cache this one too?

core/management/commands/seed.py
```diff
@@ -93,8 +93,9 @@ def seed_jhe_settings(self):
             ("site.time_zone", "string", "America/Los_Angeles"),
             ("site.registration_invite_code", "string", invite_code),
             ("auth.default_orgs", "string", ""),  # "20001:v;20002:m"
+            # IdP config is a SocialApp row (Django admin); these only gate the
+            # login button and restrict signup domains.
             ("auth.sso.saml2", "int", 0),
-            ("auth.sso.idp_metadata_url", "string", ""),
```
travis-sauer-oltech, line 97: Dropping it from the seed list won't remove it from existing databases, like UCSF or Berkeley.

Seed only creates/updates. It'll sit in the settings admin as editable but dead.

Can we add a data migration to delete it, or a note in the RFC runbook?

core/adapters.py
```diff
@@ -19,3 +22,32 @@ class JheAccountAdapter(DefaultAccountAdapter):
...
+        valid_domains = get_setting("auth.sso.valid_domains", "")
+        if not valid_domains:
+            return True
+        allowed = {d.strip().lstrip("@").lower() for d in valid_domains.split(",") if d.strip()}
+        return email.rsplit("@", 1)[-1] in allowed
+
+    def populate_user(self, request, sociallogin, data):
+        user = super().populate_user(request, sociallogin, data)
+        user.user_type = "practitioner"
+        user.identifier = sociallogin.account.uid
```
travis-sauer-oltech, line 83: `identifier` is only set on signup, so email-linked users keep their old one.

Just clarifying that this is intentional?

core/adapters.py
```diff
@@ -19,3 +22,32 @@ class JheAccountAdapter(DefaultAccountAdapter):
...
+        if not email:
+            # No email asserted (misconfigured IdP attribute mapping): fail
+            # closed rather than fall through to allauth's type-any-email
+            # signup form, which would mint an unverified practitioner.
+            return False
+        valid_domains = get_setting("auth.sso.valid_domains", "")
+        if not valid_domains:
+            return True
+        allowed = {d.strip().lstrip("@").lower() for d in valid_domains.split(",") if d.strip()}
+        return email.rsplit("@", 1)[-1] in allowed
```
travis-sauer-oltech, line 78: Can we use `partition("@")[2]` instead of rsplit here?

jhe/settings.py
```diff
@@ -107,6 +111,22 @@
 ACCOUNT_LOGIN_BY_CODE_SUPPORTS_RESEND = True
 ACCOUNT_EMAIL_CONFIRMATION_AUTHENTICATED_REDIRECT_URL = "/allauth/email/"
 ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS = 1
+# SAML SSO via allauth's saml provider. IdP config (metadata URL, attribute
+# map, SP certs) is a per-deployment SocialApp row in Django admin; the login
+# button is gated by the auth.sso.saml2 JheSetting. See docs/rfcs/0003.
+SOCIALACCOUNT_ADAPTER = "core.user_messages.JheSocialAccountAdapter"
+# "none": the IdP asserts the email (default falls through to the account
+# setting above, which would bounce SSO users through email verification).
+SOCIALACCOUNT_EMAIL_VERIFICATION = "none"
```
travis-sauer-oltech, line 121: Nothing sets `JheUser.email_is_verified` on this path, so SSO users show as unverified in the admin.

Do you think we should set it?

docs/rfcs/saml-on-allauth.md
```diff
@@ -0,0 +1,109 @@
...
+
+python3-saml signs in-process through the `xmlsec` Python binding, whose
+wheels bundle libxmlsec1/libxml2 — no `xmlsec1` binary needed. Verified in
+the built image: xmlsec and lxml both link libxml2 2.14.6 (no
+version-mismatch failure mode) and `OneLogin_Saml2_Auth` imports cleanly.
+**No Dockerfile changes.**
+
+## 5. Cons / accepted risks
+
+- allauth's socialaccount app brings its own DB tables (migrations
```
travis-sauer-oltech: Can we add the three risks I flagged in the code to this list? Signing defaults off, patients can log in via SAML, and email auth wipes the password of anyone without a verified allauth EmailAddress row.

docs/rfcs/0003-saml-on-allauth.md
```diff
@@ -0,0 +1,109 @@
...
+## 2. The change
+
+| | before (#679) | after (#697) |
+|---|---|---|
+| Package | django-saml2-auth-community (+ pysaml2, pyopenssl) | `django-allauth[saml]` (+ python3-saml, xmlsec) |
+| ACS endpoint | vendored `acs()` at `/sso/acs/` | allauth's at `/allauth/saml/<org>/acs/` |
+| IdP config | `auth.sso.idp_metadata_url` JheSetting + trigger hook | `SocialApp` row (provider `saml`) in Django admin |
+| User provisioning | vendored `get_or_create_user()` | `JheSocialAccountAdapter` (~30 lines) |
+| Login button | `auth.sso.saml2` JheSetting | flag **and** exactly one saml SocialApp (else the button hides instead of 500ing the login page) |
+| Domain restriction | `auth.sso.valid_domains` seeded but **never read** | enforced at first-time signup (not per-login; revoke = deactivate the user) |
```
travis-sauer-oltech: With AUTO_CONNECT on, anyone with an existing account skips the gate, which will be most people after the first rollout. Can we say that here?

docs/rfcs/saml-on-allauth.md
```diff
@@ -0,0 +1,109 @@
...
+- **Companion PR:** #697 (stacked on #679 / RFC 0002)
+- **Primary reviewers:** @s1monj (architecture), @travis-sauer-oltech (SAML behavior — author of the integration this replaces)
+- **Follows:** RFC 0002 §5's resolution; provenance per RFC 0001
+
+**TL;DR:** Practitioner SAML SSO moves from the vendored django_saml2_auth
+integration to allauth's built-in SAML provider (`django-allauth[saml]`) —
+the stack that already runs patient login and password reset. ~250 lines of
+vendored auth code are deleted, the pysaml2/pyopenssl chain disappears
+(retiring RFC 0002's resolver override, §6 trigger 3), and the
+missing-`xmlsec1` deployment gap closes (§4). Behavior is greenfield: SAML
```
travis-sauer-oltech: Can we restate the evidence in a line instead of pointing at 0002 Section 5?

docs/rfcs/saml-on-allauth.md
```diff
@@ -0,0 +1,109 @@
...
+decision. (Also fixed en route: `ACCOUNT_USER_MODEL_USERNAME_FIELD = None`,
+required for JHE's username-less user model on allauth's social save path.)
+
+Alternatives rejected: staying on the community fork (permanent second auth
+stack + vendored code + override forever; Simon's direction was explicit);
+dropping SAML (a product decision nobody made — allauth keeps it at near-zero
+cost); IdP config in JheSettings via a custom `get_app` (more glue for a
+setting never once populated; SocialApp admin is the native pattern).
+
+## 3. Operational runbook (first real IdP)
```
travis-sauer-oltech: Section 3 only covers turning SAML on. What's the back-out if the IdP integration goes wrong? Flipping the flag off doesn't help anyone whose password got wiped on link.

docs/rfcs/0003-saml-on-allauth.md
```diff
@@ -0,0 +1,109 @@
+# RFC 0003: SAML SSO on allauth — one auth stack, no vendored ACS, no override
+
+- **Status:** Discussion
```
travis-sauer-oltech: We already have an RFC flow in `.github/ISSUE_TEMPLATE/02_rfc.yml`, used for #334 then #391 to implement. Can we keep to that instead of a second system in docs/rfcs/? At minimum, what moves this doc from Discussion to Accepted, and who does it?

travis-sauer-oltech (changes requested):

jhe/settings.py
surfdoc: Confirmed against installed allauth 65.18: `lookup()` sets `_did_authenticate_by_email`, and `wipe_password` fires later in `_accept_login` (`flows/login.py:22`) whenever the matched user has no *verified* `EmailAddress` row — and only the patient OTP view ever writes those rows. Fixed in 12d340c with the backfill in `pre_social_login`, keyed on `_did_authenticate_by_email` (the exact field allauth keys the wipe on) and marking the row verified — which also upgrades the unverified row left behind by an abandoned OTP request, which would otherwise still trip the wipe. Pinned end-to-end by `test_acs_existing_practitioner_keeps_password`.

core/templates/registration/login.html
surfdoc: Fixed in 12d340c using your suggestion. One detail kept me from just dropping the kwarg: the tag only auto-propagates `?next=` when the kwarg is absent, but with no kwarg and no `?next=` the flow falls through to `LOGIN_REDIRECT_URL` — unset here, so Django's `/accounts/profile/` default, a 404 — hence the explicit fallback stays. allauth runs the value through `is_safe_url` before redirecting, so no open-redirect surface.

docs/rfcs/0003-saml-on-allauth.md
surfdoc: The required signing config is now documented in the operator setup guide — the SSO section of `jhe/auth.md` (jupyterhealth/software-documentation#115) shows the `advanced` block with allauth's exact snake_case keys (`authn_request_signed`, `want_assertion_signed`, plus the key/cert pair the request-signing needs). Two nuances for the record: the old `not DEBUG` pin was never operative (no KEY_FILE/CERT_FILE on main, and SAML was never configured in any deployment), and even with the flags off, python3-saml in strict mode — allauth's default — rejects any response carrying no IdP signature at all (`onelogin/saml2/response.py:288`, outside the strict-only block) and validates whichever signature is present. The flags close the response-signed-but-assertion-unsigned gap and turn on AuthnRequest signing; they're not the difference between validation and none.

core/adapters.py
surfdoc: Confirmed — `is_open_for_signup` only runs on the brand-new-account branch; email-authentication and already-linked logins never touch it. Fixed in 12d340c: `pre_social_login` (the adapter method rather than the signal — allauth documents signal ordering as undefined) raises `SignupClosedException` for existing patient accounts, on every login path including already-linked, and before `_accept_login`, so a rejected patient also keeps their password. One deliberate choice: the guard is `is_patient()` rather than not-`is_practitioner()` — `user_type` is nullable and superusers can have neither profile, so a literal non-practitioner check would lock admins out of SSO. Pinned by `test_acs_existing_patient_rejected`.

jhe/settings.py
surfdoc: The justification now lives in the settings comment at the `SOCIALACCOUNT_LOGIN_ON_GET` line and the PR description's accepted-risks section. The reasoning: the harmful login-CSRF variant — victim silently logged into an *attacker's* account — is blocked independently of `LOGIN_ON_GET`: SP-initiated responses must carry an `InResponseTo` matching a state stashed in the victim's own session (unknown state → `PermissionDenied`), and unsolicited IdP-initiated assertions are rejected by default (`reject_idp_initiated_sso`, which stays default). The residual is a third-party page force-starting the victim's *own* login; the interstitial confirm page would only add a click guarding that drive-by self-login, which we trade for one-click SSO from our button.

jhe/settings.py
surfdoc: Right on all three points (`can_authenticate_by_email` reads only the `email_authentication` app key with the global fallback; `verified_email`/IdP-asserted `email_verified` only control address-verification; and the cross-ref was wrong). Rather than rewording the comment to match a leaky global, 12d340c restores the stated design: the global `SOCIALACCOUNT_EMAIL_AUTHENTICATION = True` is gone (allauth's default False stands) and email-authentication is enabled per-IdP via `"email_authentication": true` in the SocialApp JSON — the key allauth actually consults. `AUTO_CONNECT` stays global (no per-app key exists; it's inert until email-auth fires). The settings comment, the operator guide (software-documentation#115), and the PR description now document both required keys, including the `email_verified` attribute-mapping path that grants verified treatment without any `verified_email` declaration. The PR body bullet is corrected in place.

tests/backend/test_saml_allauth.py
surfdoc: Added `SamlAcsFlowTests` in 12d340c, driving the full ACS → finish-ACS flow with a mocked python3-saml auth object (the installed wheel ships no upstream tests to copy, so signature validation stays python3-saml's job — the metadata test still exercises the real xmlsec chain): new-practitioner provisioning end-to-end, existing-practitioner email-authentication login asserting the password *survives* (per the backfill fix in the other thread), existing-patient rejection, and domain rejection rendering `account/signup_closed.html`.

docs/rfcs/0003-saml-on-allauth.md
surfdoc: Moot now, in the best way: per your process comment, the `docs/rfcs/` folder is removed from this PR entirely (and #709 removes 0002 from main), so there's no longer any doc referencing an RFC 0001. For the record, that reference pointed at the design doc shipped with the closed PR #677 — the numbering was self-consistent, just never all present in one tree.

docs/rfcs/saml-on-allauth.md
surfdoc: Mechanism confirmed, with one scoping note: because email-authentication is on and `JheUser.email` is unique, a transient NameID yields N junk SocialAccount rows for the *same* user (auto-connect persists one per login) plus one junk first `identifier` — not duplicate JheUsers. The stable-uid requirement now lives in the operator setup guide (software-documentation#115): the IdP must assert `saml-subject-id` or map `uid` to a persistent attribute, with the onboarding verification step (log in twice, confirm exactly one SocialAccount). A runtime guard isn't cleanly possible — allauth never surfaces the NameID *format* to the sociallogin — so the docs carry it.

core/adapters.py
surfdoc: Moved in 12d340c: `git mv core/user_messages.py core/adapters.py`, with the two settings dotted paths and the one test import updated — nothing else references the module (adapter paths resolve lazily, and both auth test files exercise them, so a missed reference would fail loudly).

core/context_processors.py
surfdoc: Cached in 12d340c — django cache with the same 60s TTL, importing `DEFAULT_CACHE_TTL` from `jhe_settings` so the two staleness windows stay coupled. Deliberately not the file's other pattern (`lru_cache`): that never expires within a process, so an admin adding the SAML SocialApp wouldn't see the button until every worker restarts; SocialApp rows are runtime admin config, unlike the deploy-time OIDC client id.

core/management/commands/seed.py
surfdoc: Data migration added (0045 — parked clear of #681's 0043/0044, which share the same 0042 parent; a comment in the file documents the one-line dependency flip for whichever PR merges second). Modeled on 0030's `RunPython` delete of the superseded `client.invitation_url` rows. Risk-free per your own framing: the seeded value was always `""` and nothing reads the key post-PR. Went with the migration over a docs note since migrations auto-apply on deploy — no operator action for UCSF/Berkeley. `auth.sso.saml2` and `auth.sso.valid_domains` are untouched; both remain live.

core/adapters.py
surfdoc: Intentional, for two reasons: it matches the removed django-saml2-auth path (`identifier` was only ever passed to `create_new_user`; existing users were never updated), and overwriting on link would be unsafe — `identifier` is overloaded (patients carry `ow:<id>` for the Open Wearables linkage; `get_by_ehr_id` treats it as an EHR id), so an email-linked account keeps whatever it had. The adapter docstring now states that `user_type`/`identifier` are set at first-time provisioning only (12d340c).

core/adapters.py
surfdoc: Keeping `rsplit` deliberately: the domain is what follows the *last* `@` — RFC 5321 quoted local parts may contain `@`, so `partition("@")[2]` would misparse `"user@internal"@example.com` into `internal"@example.com` and deny a legitimately allowed domain (fail-closed, so no security angle either way). It's the same parse Django's `EmailValidator` uses (`value.rsplit("@", 1)`). allauth's own `is_email_verified` does use `partition`, so there's no consistency win in either direction; added a comment at the site so the choice doesn't read as arbitrary (12d340c).

jhe/settings.py
surfdoc: Set in `populate_user` now (12d340c), with one framing note: this is cosmetic/consistency rather than a functional gap — nothing in the codebase gates on `email_is_verified` (only the legacy verify-email flow writes/reads it), allauth's own `EmailAddress` ledger is what the new stack consults and it's already correct for SSO users, and the old django-saml2-auth path didn't set the flag either. The one side effect is desirable: flipping it invalidates outstanding legacy verification links via the token hash. (Noticed while tracing, separate from this PR: `common.py`'s legacy re-send calls `request.user.send_email_verificaion` without parens, so it never fires — fixed in #708.)

docs/rfcs/saml-on-allauth.md
surfdoc: All three are now stated in the PR description's accepted-risks section, worded to track the code fixes from the other threads: signing defaults as mitigated-by-docs (the setup guide in software-documentation#115 requires the `advanced` block), and the patient login + password wipe as mitigated-by-adapter (rejection + `EmailAddress` backfill, both pinned by the new ACS tests). Also added a fourth from the email-auth thread: the IdP-asserted `email_verified` attribute-mapping path that grants verified-email treatment without the `verified_email` declaration.

docs/rfcs/0003-saml-on-allauth.md
surfdoc: Now stated in the PR description's provisioning-semantics section: existing same-email accounts log straight in via email-authentication and never pass through `is_open_for_signup`, so the `auth.sso.valid_domains` gate applies only to users JHE has never seen — the minority after first rollout. One attribution tweak from your phrasing: the bypass is caused by email-authentication (the by-email lookup that makes the login `is_existing` before the signup branch), not AUTO_CONNECT, which only persists the link. The adapter docstring says the same.

docs/rfcs/saml-on-allauth.md
surfdoc: Overtaken by the docs/rfcs removal — the TL;DR (and the pointer you flagged) went away with the doc. The evidence itself is on the record in #679's design writeup §5: verified live on fly.io 2026-07-31 — no `xmlsec1` binary in the image, `auth.sso.saml2 = 0`, no IdP metadata ever configured.

docs/rfcs/saml-on-allauth.md
surfdoc: Back-out is now a "Disabling SAML SSO (back-out)" section in the operator guide (software-documentation#115): flag-off only hides the button (allauth's URLs stay mounted); deleting the Social Application is the real kill switch (every `/allauth/saml/<slug>/*` request then 404s); optional SocialAccount cleanup for a permanent back-out, since leftover rows would silently re-link users on a future re-enable. On the password half: with the `pre_social_login` backfill from the other thread, linking no longer wipes anything — and the guide documents that a user without a usable password must use allauth's `/allauth/password/reset/`, because the login page's "Forgot Password" (Django's `PasswordResetForm`) silently skips exactly those users.

docs/rfcs/0003-saml-on-allauth.md
surfdoc: I backed out this RFC flow. My apologies, I am not familiar with how RFC worked for this repo and that we had an rfc.yml. I have always hashed things out in Slack and then built it. I misinterpreted a Slack message from Simon and thought he wanted RFCs for the Capabilities Statement and other work so included it in this as well but probably not necessary now since the code is already in a PR state.

docs/rfcs/saml-on-allauth.md
surfdoc: Updated for the consolidation: the RFC-index half is moot — the `docs/rfcs/` folder is removed from this branch entirely (and #709 removes 0002 from main), so there's nothing left to link. The software-documentation half stands: all three staleness points checked out and jupyterhealth/software-documentation#115 rewrites the SSO section for the SocialApp-based config and the `/allauth/saml/<slug>/acs|metadata/` URLs — now also carrying the stable-uid requirement and the back-out procedure that previously lived in the RFC doc. One correction it also makes: the Mock SAML Audience must be the SP *entity ID*, which allauth defaults to the metadata URL, not the ACS URL. Sequenced to merge when this PR lands.

core/context_processors.py
```diff
@@ -28,6 +30,23 @@ def _get_oidc_client_id():
...
+    # of which would 500 the whole login page). Hide the button instead;
+    # multi-IdP deployments need per-IdP buttons in a customized template.
+    if not get_setting("auth.sso.saml2", 0):
+        return False
+    # Same TTL cache as get_setting (not lru_cache: SocialApp rows are runtime
+    # admin config, so the two gates must go stale together, within 60s).
+    cache_key = "saml2_single_social_app"
+    enabled = cache.get(cache_key)
+    if enabled is None:
+        enabled = SocialApp.objects.filter(provider="saml").count() == 1
```
travis-sauer-oltech, line 49: Can we clear the key from a post_save/post_delete receiver on SocialApp in core/signals.py?

This is the back-out step the setup guide documents.

core/adapters.py
```diff
@@ -0,0 +1,87 @@
...
+        email = getattr(sociallogin, "_did_authenticate_by_email", None)
+        if email and sociallogin.user.pk:
+            address, _ = EmailAddress.objects.get_or_create(
+                user=sociallogin.user,
+                email=email.lower(),
+                defaults={"verified": True, "primary": True},
+            )
+            if not address.verified:
+                address.verified = True
+                address.save(update_fields=["verified"])
```
travis-sauer-oltech, line 62: This marks the allauth row verified but leaves `JheUser.email_is_verified` = `False`.

`populate_user` only reaches brand-new SSO users. therefore, an email-linked practitioner will show unverified from JHE admin panel.

Can we set it here too?

tests/backend/test_saml_allauth.py
```diff
@@ -0,0 +1,215 @@
...
+        _set_setting("auth.sso.saml2", "int", 0)
+        response = self.client.get("/accounts/login/")
+        self.assertNotContains(response, "Continue with SAML2")
+
+    def test_button_links_to_allauth_when_enabled(self):
+        SocialApp.objects.create(provider="saml", name="Test IdP", client_id="idp")
+        _set_setting("auth.sso.saml2", "int", 1)
+        response = self.client.get("/accounts/login/")
+        self.assertContains(response, "Continue with SAML2")
+        self.assertContains(response, "/allauth/saml/idp/login/")
```
travis-sauer-oltech, line 70: The href assertion stops before the query string, so we don't know if the value after next carries through.

Can we assert the default next is in the href, plus a case with ?next= set on the login page?

tests/backend/test_saml_allauth.py
```diff
@@ -0,0 +1,215 @@
...
+    def test_acs_provisions_new_practitioner(self):
+        self._post_acs("newdoc@example.org")
+        user = JheUser.objects.get(email="newdoc@example.org")
+        self.assertEqual(str(user.pk), self._logged_in_user_id())
+        self.assertEqual(user.user_type, "practitioner")
+        self.assertTrue(Practitioner.objects.filter(jhe_user=user).exists())
+        self.assertTrue(user.email_is_verified)
+        self.assertEqual(SocialAccount.objects.get(user=user, provider="saml").uid, "idp-uid-1")
+
+    def test_acs_existing_practitioner_keeps_password(self):
```
travis-sauer-oltech, line 197: Can we fadd a test that sets is_active = False on a practitioner, posts to ACS, and asserts they are not logged in?

tests/backend/test_get_setting_migration.py
travis-sauer-oltech: `SAML2_ENABLED` is now the bool from _saml2_enabled, not the int from the DB.

Comment is wrong now.

Can we say bool and assert False?

travis-sauer-oltech (changes requested):
@surfdoc After these changes, please feel free to merge. I resolved all of the conversations that you addressed from my first review. Thanks!

core/adapters.py
surfdoc: Fixed in b34ab0e. `pre_social_login` now sets `user.email_is_verified = True` when it backfills the verified `EmailAddress` for an email-linked user, so a practitioner reached through email-authentication is no longer left unverified in the admin. `populate_user` already covered brand-new SSO users.

core/context_processors.py
surfdoc: Done in b34ab0e. Added `post_save`/`post_delete` receivers on `SocialApp` in `core/signals.py` that clear the `saml2_single_social_app` cache key, so adding or removing an IdP row flips the login button immediately — the setup guide's back-out step — instead of waiting out the TTL. The key is now a `SAML2_SINGLE_SOCIAL_APP_CACHE_KEY` constant in `context_processors.py`, shared by both sides.

tests/backend/test_saml_allauth.py
surfdoc: Added in b34ab0e. `test_button_links_to_allauth_when_enabled` now asserts the default destination is carried through the href (`next=%2Fclients%2Fjhe-admin%2Forganizations`), and a new `test_button_forwards_login_page_next` sets `?next=` on the login page and asserts it reaches the button href — so SSO lands where a same-page password login would.

tests/backend/test_saml_allauth.py
surfdoc: Added `test_acs_inactive_practitioner_not_logged_in` in b34ab0e: an `is_active=False` practitioner posts to ACS and we assert no session is created. It passes — allauth blocks the inactive user at login, which confirms the adapter's documented "deactivate the user to revoke access".

tests/backend/test_get_setting_migration.py
surfdoc: Fixed in b34ab0e. The comment now says the value is the bool from `_saml2_enabled()`, and the assertion is `assertIs(ctx["SAML2_ENABLED"], False)` — strict identity, so a regression back to the raw DB int would fail.

PR #745, opened 2026-08-29
fix: unbreak lowercase `/fhir/r5/` prefix, which is still in active use

minrk:
unbreaks CareX measurement upload, for example


minrk:
done in #748

PR #760, opened 2026-09-01
feat: ow_poll ingests blood glucose and follows OW pagination

travis-sauer-oltech:
Polls blood glucose in addition to heart rate, driven by the patient's consented scopes. Also follows OW's pagination cursor, which was capping each run at 50 samples.

Needs omh-shim 1.5.0.


surfdoc (approved):
LGTM 🚀

Issue #731, opened 2026-08-23
[RFC]: Redesign the EHR Patient Portal client UI with a themeable brand layer


surfdoc:
## Problem
The EHR Patient Portal client (`clients/ehr-patient-portal/`, issue #489; renamed from "Patient Access" in #744 to avoid colliding with the patient sign-in-code client) works but is visually bare: an unstyled heading, a single Bootstrap search box, and a raw `<pre>` element that appends status text line by line. There is no branding, no sense of progress, and no styled success or error state. For a patient-facing flow that asks people to connect their medical records, this reads as unfinished and does little to build trust. The UI is also implicitly styled for no one in particular, so an open-source adopter cannot easily apply their own identity.

## Goal
A polished, trustworthy, accessible patient flow that ships with JupyterHealth's identity by default, and that any deployer can rebrand by changing a small set of values (colors, logo, product name) without editing templates. Stay inside the existing stack (Django templates plus the already-vendored Bootstrap 5 and vanilla JS); no new frontend framework or build step. Keep the change scoped to the patient-facing flow, so the admin portal is untouched.

## Current Approach
Two Django-rendered pages, `connect.html` and `callback.html`, extend a shared `base.html` that loads Bootstrap and the SMART `fhir-client.js`. `client-ehr-patient-portal.js` runs the flow (redeem invitation, get a JHE token, pick an organization, authorize with Epic via SMART, pull records, convert R4 to R5, write to JHE) and reports progress by appending text to a `<pre>`. There is no theme layer and no brand configuration.

## Proposal
Restyle the EHR Patient Portal templates and add one scoped theme stylesheet whose entire palette, type, and radii are CSS custom properties. That token file is the single surface an adopter overrides to rebrand. Brand specifics (logo, product name) are supplied from JHE Settings with JupyterHealth defaults, injected through the existing context processor (see the resolution of the open questions below); the organization name is shown dynamically from the picked brand across all screens. Visible changes: a branded header, a three-step progress rail (Choose organization, Sign in, Import records), a real organization picker, and the `<pre>` log replaced by a live import ledger with per-category counts, a success/receipt screen, and proper error states. The UI font (Inter, matching jupyterhealth.org) is vendored locally to work under strict CSP. No framework change; the unused `patient-frontend` React bundle was removed separately in #733.

Mockups follow (an interactive prototype was reviewed separately; the last image shows a full rebrand achieved by changing only tokens, logo, and name).

**Connect: choose your organization**

[...]

PR #765, opened 2026-09-04
feat(patient-ui): themeable, branded patient-facing journey (RFC #731)

surfdoc:
## Summary

This builds the patient journey from RFC #731. A patient opens their invitation link, chooses what to share, consents, connects their device or hospital, watches the import, and can later stop sharing.

It follows the admin UI's architecture. Each client has one thin Django shell page. The screens are Handlebars components under `common/patient_facing/components/`, rolled up per page the way `clients/jhe_admin/portal.html` does. The logic is vanilla JavaScript in `common/js/patient-facing.js`, in the style of `client-jhe-admin.js`. Consent is written only through the existing `patients/{id}/consents` API.

### How it works

**Entry and auth are unchanged.** The invitation link is the client's own page (`/clients/ow/launch?code=…` or `/clients/ehr-patient-portal/?code=…`). `patientApp()` redeems the code at `/o/token/`, keeps the bearer token in `sessionStorage`, and reads the patient id from `users/profile`. There is no Django session.

**One invitation is one client.** The view builds `PATIENT_PORTAL_CONFIG` from the JheClient's `aux_data` (Epic client id and SMART scopes) and its `ClientDataSource` rows. The hub lists only those data sources. There is no cross-client routing.

**The shared app** in `common/js/patient-facing.js` provides:
- a `?route=` router with browser history
- `pfRender`, which renders a Handlebars component into `#pf_main`
- `pfApi`, a bearer-token fetch wrapper
- `pfSources`, which turns the consents payload into a per-source view model (consented and pending scopes, labels, studies)
- the hub, consent, done, manage, importing and error screens

**Client hooks.** `client-ow.js` and `client-ehr-patient-portal.js` each register `pfClient.connect(source)`. For Oura that is the launch card, then the Open Wearables user, then Oura authorize. For Epic it is the hospital picker, then the SMART launch. The Epic callback page renders the importing screen, runs the existing import, and hands off to `?route=done`.

**Consent and revoke** go through `PatientViewSet.consents`. Never-asked scopes are POSTed; revoked rows are PATCHed. "Stop sharing" PATCHes `consented=false`, which runs the Open Wearables disconnect hook. Nothing is deleted.

**One backend addition.** `FhirSourceSerializer` gains read-only `resource_counts` and `facility`, so the receipt and the source cards can show per-type counts and the chosen facility.

**Everything else on the server is untouched.** `PatientViewSet.consents`, `PatientInvitation`, the Open Wearables proxy endpoints and every route on `main` are as they were. `clients/ow/` and `clients/ow/complete` keep their old names as aliases. The legacy `clients/ow/manage` page is removed; the manage screen replaces it. The brands search now also matches state names and ZIP prefixes.

### Notes for reviewers

- **Three deliberate departures from the RFC spec.** `showFlowError` lives in `patient-facing.js` rather than `common.js`, which the admin also loads. `PATIENT_PORTAL_CONFIG` carries `expectedResourceTypes`, `siteTitle` and `pageUrl` beyond the spec's five keys, because the receipt's "Not synced" list, the launch card copy and the error screen's Back link need them. `facility` rides along with `resource_counts` so the cards can name the chosen hospital.
- **Connect-time consent is enforced in the browser**, as on `main`: the connect route sends an unconsented source to the consent screen first. The API's permission checks and the Open Wearables disconnect hook remain the server-side guarantees.
- **Consent has no data-source dimension.** Badges are derived per source through `StudyDataSource`. A new study requesting an already-consented scope shows the source as Not consented until the patient re-agrees. Connection status is a separate RFC.
- **Shared globals in `patient-facing.js` carry a `pf` prefix** because the file shares the page's global scope with a client script.
- **Components are rolled up in `body_footer`, not `html_footer`.** The admin portal includes its components in `html_footer`, which the shared base places after `</body>`. `body_footer` sits inside the body, after Handlebars has loaded, so the `<script type="text/template">` tags are valid HTML instead of content the parser reparents into the body. The auth pages (`login`, `callback`, `callback_popup`) already use `body_footer`, so it is the majority pattern in the shared base.
- **Cache-busting stays `?v=JHE_VERSION`.** Whitenoise caps browser caching at 60 seconds anyway. Manifest storage is a project-wide change to file separately.
- **No migration.**

### How to review

Seven self-contained commits on current `main`, best read in order:

1. **Branding layer:** tokens stylesheet, vendored Inter, lockup, header partial, invitation email, `site.ui.*` settings. No behaviour.
2. **Shared app and components:** `patient-facing.js` and the Handlebars components. This is the substance of the change.
3. **Client shells and hooks:** the two client pages become shells, the client scripts shrink to their connect hooks, `views/patient_facing.py` builds the config.
4. **API:** `resource_counts` and `facility` on the FhirSource serializer.
5. **Seed:** settings entries and the demo study's wildcard scope.
6. **Test:** the Open Wearables disconnect on revoke.
7. **JS syntax:** `patient-facing.js` and the two client hooks move from `var` to `const`/`let`, arrow callbacks and template literals, matching `client-jhe-admin.js`. The pull-list test in `test_fhir_r4_import.py` reads the `const` declaration. No behaviour change.

### Testing

- Backend: 1168 passed. New tests cover shell config injection, `resource_counts`, seed URLs, and the Open Wearables disconnect on revoke.
- Jest: 107 passed across 8 suites. New: `pfSources`, routing and token redeem, consent writes, receipt, one render test per component. Updated: the EHR picker and import suites.
- `ruff check` and `ruff format --check` clean.
- Walked both journeys locally. Oura: to the Oura login page and back through the vendor-error path. Epic sandbox: import of 298 records to the receipt, then manage and revoke. Reload and browser Back on every route. Malformed, expired and reused invitation codes.
- Re-walked after the syntax pass: invitation redeem, hub, done receipt and manage on both clients in Chrome, with no console errors.

### Note for @s1monj

The seeded Lifespan BP & HR study now requests the wildcard FHIR scope (`http://hl7.org/fhir/resource-types` / `*`) through the EHR Patient Portal data source, so the demo's EHR invitation opens on a hub with something to consent to. The scope keeps its "All FHIR Resources" label in the seed and admin UI. The patient screens show it as "Clinical records".

Design decisions are recorded in the ADR on #731.

Ref: RFC #731.


s1monj:
LGTM - a couple of small refactors to keep terminology consistent
- renamed PATIENT_PORTAL_CONFIG to PATIENT_FACING_CONFIG
- renamed facility to ehr_brand_location_name

